PR #38, opened 2020-11-18
Added support for new session json for each user

dalmia:


dalmia:
Seems straightforward -> not writing comments

src/views/Player.vue
```diff
@@ -97,7 +100,8 @@ export default {
           process.env.VUE_APP_BACKEND +
             process.env.VUE_APP_BACKEND_PLIO_DETAILS +
             "?plioId=" +
-            this.$route.params.id
+            this.$route.params.id +
+            "&userId=" + this.studentId
```
pritamps, line 104: Why this mismatch between userID and studentID? make it userID only everywhere no?

src/views/Player.vue
```diff
@@ -97,7 +100,8 @@ export default {
           process.env.VUE_APP_BACKEND +
             process.env.VUE_APP_BACKEND_PLIO_DETAILS +
             "?plioId=" +
-            this.$route.params.id
+            this.$route.params.id +
```
pritamps, line 103: oh.... this is needed urgently no? so is okay. We can merge.

Because getting video details shouldn't depend on user ID (except for authentication). Abhi ke liye merge kar lo. Let me think a bit. With current JSONs this maybe fine, but in near future, we will change in some way.

Just to keep in mind

pritamps (changes requested):
Only one change. Change `studentId` to `userId` everywhere for consistency

src/views/Player.vue
```diff
@@ -61,7 +61,8 @@ export default {
         'failsafeUrl': ''
       },
       enterFullscreen: [],
-      exitFullscreen: []
+      exitFullscreen: [],
+      sessionId: 1
```
pritamps, line 65: `sessionId` is always one? Or is there somewhere you are incrementing if file exists already on backend (didn't see in that PR)?

src/views/Player.vue
```diff
@@ -106,6 +110,7 @@ export default {
           this.plioId = res.data.plioId;
           this.browserErrorHandlingValue.failsafeUrl = res.data.plioDetails.failsafe;
           this.isFullscreen = false;
+          this.sessionId = res.data.sessionId;
```
dalmia, line 113: @pritamps `sessionId` is updated here.

src/views/Player.vue
dalmia: It is getting updated. Tagged you in a comment pointing towards the code change.

src/views/Player.vue
dalmia: Yep. And the `userId` param is optional in the backend code - One can send in the `userId` here as well to get the session ID - or else if it is not passed - sessionId will be 1. But I agree, this can be a different API - separate from the one supposed to only get the plio details.

PR #39, opened 2020-11-20
MRG: Frontend changes to adhere to new format

dalmia:
- [x] remove options from response JSON
- [x] added `userId`, `sessionId` and `plio_id` in `response` - deleted `meta`. 
- [x] updated backend code ([PR](https://github.com/avantifellows/plio-backend/pull/14))
- [x] using a single `journey` variable to capture all events
- [x] disabled check for browser support for browsers we know work
- [x] log option click

### DIFF
--- a/src/components/PlioQuestion.vue
+++ b/src/components/PlioQuestion.vue
@@ -28,7 +28,7 @@
             <ul>
               <li class="option">
                 <div
-                  v-for="option in plioQuestion.item.question.options"
+                  v-for="(option, index) in plioQuestion.item.question.options"
                   :key="option"
                   class="answer_option radio"
                   :ref="option"
@@ -41,6 +41,7 @@
                     name="options"
                     v-model="selectedOption"
                     :value="option"
+                    @click="this.selectOption(index)"
                   />{{ option }}
                 </label>
                 </div>
@@ -84,7 +85,7 @@
 
 <script>
 export default {
-  name: "IvideoQuestion",
+  name: "PlioQuestion",
   props: ["plioQuestion"],
   data() {
     return {
@@ -115,6 +116,13 @@ export default {
     closeModal() {
       this.show = false;
       document.querySelector("body").classList.remove("overflow-hidden");
+    },
+
+    selectOption(option_index) {
+        this.$emit('update-journey', 'option-selected', {
+            'question': Number(this.plioQuestion.id),
+            'option': option_index
+        });
     },
 
     // Opens the question window
--- a/src/components/PlioQuestion.vue
+++ b/src/components/PlioQuestion.vue
@@ -41,7 +41,7 @@
                     name="options"
                     v-model="selectedOption"
                     :value="option"
-                    @click="this.selectOption(index)"
+                    @click="selectOption(index)"
                   />{{ option }}
                 </label>
                 </div>
--- a/src/views/Player.vue
+++ b/src/views/Player.vue
@@ -215,13 +215,13 @@ export default {
           .catch(err => console.log(err))
     },
 
-    updateJourney(event, details = {}) {
+    updateJourney(logEvent, details = {}) {
         // handle the case when fullscreen has been clicked but Plyr has not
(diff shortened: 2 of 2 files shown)


src/views/Player.vue
```diff
@@ -64,11 +63,10 @@ export default {
...
         'failsafeUrl': ''
       },
-      fullscreenJourney: [],
-      questionJourney: [],
-      playbackJourney: [],
+      journey: [],
       hasVideoPlayed: -1,
-      sessionId: 1
+      sessionId: 1,
+      hasPlyrLoaded: false
```
dalmia, line 69: checks if `plyr` has loaded or not - it is needed as if someone presses fullscreen before plyr has loaded, currentTime is NaN

src/views/Player.vue
```diff
@@ -64,11 +63,10 @@ export default {
         'failsafeType': 'g-form',
         'failsafeUrl': ''
       },
-      fullscreenJourney: [],
-      questionJourney: [],
-      playbackJourney: [],
+      journey: [],
```
dalmia, line 66: this will capture the journey now

src/views/Player.vue
```diff
@@ -147,10 +144,7 @@ export default {
             this.answers.push(plioQuestion.user_answer)
           }
 
-          // set the global list of
-          // options and time values
-          // (this.options, this.times) 
-          this.options = res.data.options
```
dalmia, line 153: removing `options` from the uploaded response now.

src/views/Player.vue
```diff
@@ -223,26 +212,17 @@ export default {
...
+        // handle the case when fullscreen has been clicked but Plyr has not
+        // yet loaded -> this.player.currentTime = NaN
+        var player_time = this.hasPlyrLoaded ? this.player.currentTime : 0
+
+        this.journey.push({
+            'event': event,
+            'details': details,
+            'system_time': String(this.getCurrentDateTime()),
+            'player_time': String(player_time)
+        })
```
dalmia, line 225: standard format for ALL events

src/views/Player.vue
```diff
@@ -264,19 +244,10 @@ export default {
...
-      // this.questionJourney.push(
-      //   [
-      //     String(this.getCurrentDateTime()),
-      //     "submit",
-      //     String("q-" + currQuesIndex)
-      //   ]
-      // )
+          "question-submitted", {
+            'question': currQuesIndex,
+            'option': plioQuestion.item.question.options.indexOf(answer) + 1
```
dalmia, line 249: for submit response, log the question number and option selected as details.

src/views/Player.vue
deepansh96: I also had a doubt when I made a change in my PR, will "event" here work properly? JavaScript reads the word "event" not as a variable, but as an event. Do you think we should avoid using the word event here?

src/views/Player.vue
deepansh96: indexOf might not work in older browsers!

src/views/Player.vue
dalmia: achcha, haan I'll change

src/views/Player.vue
deepansh96: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/indexOf#Polyfill

src/components/PlioQuestion.vue
```diff
@@ -41,6 +41,7 @@
                     name="options"
                     v-model="selectedOption"
                     :value="option"
+                    @click="this.selectOption(index)"
```
deepansh96, line 44: no need to type "this" here
this code will work
`@click="selectOption(index)"`

pritamps (approved):
This looks fine to me. But I think now I have lost track of all the changes and things we are looking out for. I have a gut feeling that there are too many checks for too many things and we can simplify (maybe even the UX) somewhat. But will have to think.

Till then, this is fine!

PR #40, opened 2020-11-20
Options persist | Marker Color Change | Done Button

deepansh96:
- [x] Done button to close the modal
- [x] Separated methods for buttons on modal
- [x] Previously answered options shown when returned to question
- [x] Tooltip color changed to green when question done

### DIFF
--- a/src/components/PlioQuestion.vue
+++ b/src/components/PlioQuestion.vue
@@ -76,7 +76,7 @@
 
           <!-- close button -->
           <button
-            v-if="isAnswerSubmitted"
+            v-else
             class="btn close"
             @click="clickClose"
           >
--- a/src/components/PlioQuestion.vue
+++ b/src/components/PlioQuestion.vue
@@ -14,10 +14,7 @@
             <div
               class="close-container"
               id="skip-button"
-              @click="
-                closeModal();
-                this.$emit('answer-skipped', this.plioQuestion);
-              "
+              @click="clickSkip"
             >
               <div class="leftright"></div>
               <div class="rightleft"></div>
@@ -51,14 +48,12 @@
           </div>
         </div>
 
+        <!-- revise button -->
         <div class="modal__footer">
           <button
             id="revise-button"
             class="btn revise"
-            @click="
-              closeModal();
-              this.$emit('revision-needed', this.plioQuestion);
-            "
+            @click="clickRevise"
           >
             Revise
           </button>
@@ -69,12 +64,23 @@
           <i class="fas fa-times-circle" ref="wrong-icon"
             v-if="isAnswerSubmitted && !isAnswerCorrect"></i>
 
+          <!-- submit button -->
           <button
+            v-if="!isAnswerSubmitted"
             class="btn submit"
             :disabled="isDisabled"
             @click="clickSubmit"
           >
             Submit
+          </button>
+
+          <!-- close button -->
+          <button
+            v-if="isAnswerSubmitted"
+            class="btn close"
+            @click="clickClose"
+          >
+            Close
           </button>
         </div>
       </div>
@@ -108,6 +114,10 @@ export default {
     // Returns the text of the correct answer
     correctAnswer() {
       return this.plioQuestion.item.question.options[this.correctAnswerIndex];
+    },
+    // Returns the index of the question that has popped up.
+    currentQuestionIndex() {
+      return this.plioQuestion.id;
     }
   },
   methods: {
@@ -119,6 +129,14 @@ export default {
 
     // Opens the question window
     openModal() {
+      // Show highlighted options if coming back to an answered question
+      // Wait 200 ms because it takes some time to find the DOM elements
+      if (this.plioQuestion.state == "answered") {
+        setTimeout(() => {
+          this.showResult();
+        }, 200);
+      }
+      
       this.text = "";
       this.show = true;
       document.querySelector("body").classList.add("overflow-hidden");
@@ -131,14 +149,39 @@ export default {
     
     // Highlights the correct option as green, wrong one as red
     showResult(){
-      if (!this.isAnswerCorrect) 
-        this.$refs[this.selectedOption].className = "answer_option-wrongAnswer"
+      if (!this.isAnswerCorrect) {
+        this.$refs[this.selectedOption].classList.add("wrongAnswer")
+      }
       
-      this.$refs[this.correctAnswer].className = "answer_option-correctAnswer"
+      this.$refs[this.correctAnswer].classList.add("correctAnswer")
+    },
+
+    removeOptionHighlight(){
+      var allOptions = document.querySelectorAll('.answer_option')
+
+      allOptions.forEach( (option) => {
+        option.removeAttribute("class");
+        option.className = "answer_option"
+      });
+    },
+
+    toggleMarkers(index){
+      var markers = document.querySelectorAll('#marker')
+      
+      if (markers[index].className == 'tooltip'){
+        markers[index].classList.remove('tooltip')
+        markers[index].classList.add('tooltip-answered')
+      }
     },
 
     // Things to do after submit is clicked
     clickSubmit(){
+
+      // On clicking submit, previous highlights should be removed
+      // so new highlights can be configured
+      this.removeOptionHighlight();
+      this.toggleMarkers(this.currentQuestionIndex)
+
       this.isAnswerSubmitted = true;
 
       document.getElementById('skip-button').hidden = true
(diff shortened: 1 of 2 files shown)


src/components/PlioQuestion.vue
```diff
@@ -14,10 +14,7 @@
             <div
               class="close-container"
               id="skip-button"
-              @click="
-                closeModal();
-                this.$emit('answer-skipped', this.plioQuestion);
-              "
+              @click="clickSkip"
```
deepansh96, line 17: consolidated into one method

src/components/PlioQuestion.vue
```diff
@@ -51,14 +48,12 @@
...
+        <!-- revise button -->
         <div class="modal__footer">
           <button
             id="revise-button"
             class="btn revise"
-            @click="
-              closeModal();
-              this.$emit('revision-needed', this.plioQuestion);
-            "
+            @click="clickRevise"
```
deepansh96, line 56: consolidated into one method

src/components/PlioQuestion.vue
```diff
@@ -69,13 +64,24 @@
           <i class="fas fa-times-circle" ref="wrong-icon"
             v-if="isAnswerSubmitted && !isAnswerCorrect"></i>
 
+          <!-- submit button -->
           <button
+            v-if="!isAnswerSubmitted"
```
deepansh96, line 69: submit button is shown only when the answer has not been submitted yet.
If answer is submitted, replace it with done button

src/components/PlioQuestion.vue
```diff
@@ -69,13 +64,24 @@
...
           </button>
+
+          <!-- close button -->
+          <button
+            v-if="isAnswerSubmitted"
+            class="btn close"
+            @click="clickClose"
+          >
+            Close
+          </button>
```
deepansh96, line 84: Added close button. Will appear when answer is submitted

src/components/PlioQuestion.vue
```diff
@@ -108,6 +114,10 @@ export default {
     // Returns the text of the correct answer
     correctAnswer() {
       return this.plioQuestion.item.question.options[this.correctAnswerIndex];
+    },
+    // Returns the index of the question that has popped up.
+    currentQuestionIndex() {
+      return this.plioQuestion.id;
```
deepansh96, line 120: will return the index of the current popped up question

src/components/PlioQuestion.vue
```diff
@@ -119,6 +129,14 @@ export default {
...
     // Opens the question window
     openModal() {
+      // Show highlighted options if coming back to an answered question
+      // Wait 200 ms because it takes some time to find the DOM elements
+      if (this.plioQuestion.state == "answered") {
+        setTimeout(() => {
+          this.showResult();
+        }, 200);
+      }
+
```
deepansh96, line 139: When question pops up + and the user is returning to a previously answered question,
need to show what was the result at that time.

Settimeout because it takes about 100-200 ms (doesn't happen instantly) for JS to find the DOM elements.

src/components/PlioQuestion.vue
```diff
@@ -131,28 +149,65 @@ export default {
...
     // Highlights the correct option as green, wrong one as red
     showResult(){
-      if (!this.isAnswerCorrect) 
-        this.$refs[this.selectedOption].className = "answer_option-wrongAnswer"
+      if (!this.isAnswerCorrect) {
+        this.$refs[this.selectedOption].classList.add("wrongAnswer")
+      }
+      
+      this.$refs[this.correctAnswer].classList.add("correctAnswer")
+    },
```
deepansh96, line 157: Broke down class
 `answer_option-wrongAnswer` into two classes => `answer_option` and `wrongAnswer`
and
 `answer_option-correctAnswer` into two classes => `answer_option` and `correctAnswer`

src/components/PlioQuestion.vue
```diff
@@ -131,28 +149,65 @@ export default {
...
+
+    removeOptionHighlight(){
+      var allOptions = document.querySelectorAll('.answer_option')
+
+      allOptions.forEach( (option) => {
+        option.removeAttribute("class");
+        option.className = "answer_option"
+      });
+    },
+
```
deepansh96, line 167: user returns to a previously answered question ->
sees that his old response is highlighted ->
chooses another option and clicks submit ->
**new answers need to be highlighted**

For the last step, we need to remove existing highlighting, if there is any

src/components/PlioQuestion.vue
```diff
@@ -131,28 +149,65 @@ export default {
...
+    toggleMarkers(index){
+      var markers = document.querySelectorAll('#marker')
       
-      this.$refs[this.correctAnswer].className = "answer_option-correctAnswer"
+      if (markers[index].className == 'tooltip'){
+        markers[index].classList.remove('tooltip')
+        markers[index].classList.add('tooltip-answered')
+      }
```
deepansh96, line 174: Color the tooltip if question has been answered

src/components/PlioQuestion.vue
```diff
@@ -131,28 +149,65 @@ export default {
...
     },
 
     // Things to do after submit is clicked
     clickSubmit(){
+
+      // On clicking submit, previous highlights should be removed
+      // so new highlights can be configured
+      this.removeOptionHighlight();
+      this.toggleMarkers(this.currentQuestionIndex)
+
```
deepansh96, line 184: Just calling the above defined methods

src/components/PlioQuestion.vue
```diff
@@ -131,28 +149,65 @@ export default {
...
+    // Things to do when revise button is clicked
+    clickRevise(){
+      this.closeModal();
+      this.$emit('revision-needed', this.plioQuestion);
+    },
+
+    clickSkip(){
+      this.closeModal();
+      this.$emit('answer-skipped', this.plioQuestion);
```
deepansh96, line 210: instead of waiting for 3 seconds ->

clickClose() method is called

src/components/PlioQuestion.vue
```diff
@@ -239,23 +294,11 @@ li {
...
-.answer_option-correctAnswer {
-  text-align: left;
-  padding: 2px;
-  margin: 5px;
-  font-size: 1.3rem;
-  margin-right: 57px;
-  border-radius: 5px;
+.correctAnswer {
   background-color:lightgreen
 }
```
deepansh96, line 299: .correctAnswer is added to the classlist of the required element -> it means that just the property we need is added

src/views/Player.vue
```diff
@@ -40,6 +40,8 @@ var timeout = null;
 
 // wait this much time (secs) then show error page
 // if browser is not supported
+
+// TODO - set back to 10 or 5 for PROD
```
deepansh96, line 44: ignore

src/views/Player.vue
```diff
@@ -393,6 +395,7 @@ export default {
           var question = plioQuestion.item;
           // Add marker to progress bar
           var marker = document.createElement("SPAN");
+          marker.setAttribute("id", "marker")
```
deepansh96, line 398: Set all markers with an id, easy to access the bunch

src/views/Player.vue
```diff
@@ -521,13 +524,14 @@ export default {
...
             if (modal != undefined)   {
+              this.player.pause();
+
               document
                 .getElementsByClassName("plyr")[0]
                 .appendChild(document.getElementsByClassName("modal")[0]);
 
               if (plioQuestion["state"] == "notshown") plioQuestion["state"] = "unanswered"; 
                 //var marker = plioQuestion["marker"];
-                this.player.pause();
```
deepansh96, line 530: moving player.pause() upwards because sometimes the player keeps playing even when modal is visible -> but i didn't see any visible change

src/components/PlioQuestion.vue
```diff
@@ -69,13 +64,24 @@
...
             class="btn submit"
             :disabled="isDisabled"
             @click="clickSubmit"
           >
             Submit
           </button>
+
+          <!-- close button -->
+          <button
+            v-if="isAnswerSubmitted"
```
pritamps, line 79: Maybe use `v-else` here?

src/components/PlioQuestion.vue
pritamps: see if you can replace the `timeout` with a listener AND a timeout? 

Something like (but maybe not the same) as https://stackoverflow.com/questions/16149431/make-function-wait-until-element-exists

pritamps (changes requested):
Very minor changes! Feel free to ignore also :P

PR #41, opened 2020-11-29
ADD: YT embed if browser error, prominent gform link

deepansh96:
image
------------------------------------------------

image

This is how the error page will look like.

We realized that we were giving a failsafe g-form link but we were not giving the youtube video link.
Added a youtube embed here and made the g-form link such that it's easily visible.

### DIFF
--- a/src/views/Error.vue
+++ b/src/views/Error.vue
@@ -14,6 +14,9 @@
 
       <!-- TODO
       remove hardcoded link -> either parameterize it, or enter a link to youtube -->
+      <div class="lead_text">
+        <p>Plio को कैसे इस्तेमाल करना है, जानने के लिए ये video देखें </p>
+      </div>
       <div class='embed-container'>
         <iframe src='https://www.youtube.com/embed/FLOwzot27XM' frameborder='0' allowfullscreen>
         </iframe>
@@ -22,15 +25,24 @@
       <!-- failsafe begins -->
       <div v-if="hasFailSafe">
 
+        <!-- TODO - Adding loading spinner till YT video loads -->
         <!-- failsafe G-form begins -->
-        <div v-if="isFailSafeGform">
+        <div v-if="isFailSafeGform && isVideoIdFetched" >
+          <hr class="solid">
+
           <div class="lead_text">
-            <p>इसी प्रतियोगिता में Google फॉर्म से हिस्सा लेने के लिए नीचे बटन पे क्लिक करें</p>
+            <p>इसी प्रतियोगिता में Google फॉर्म से हिस्सा लेने के लिए नीचे दिया गया वीडियो देखें और उसके बाद Google फॉर्म का लिंक क्लिक करें </p>
+            <div class='embed-container'>
+              <iframe :src="this.value['youtubeId']" id="lesson-video" frameborder='0' allowfullscreen></iframe>
+            </div>
+            <br>
+            <i class="far fa-hand-point-right"></i>
             <a
-              :href=value.failsafeUrl
-              class="icon-block">
-              <img src="../assets/google_form.svg">
+            :href=value.failsafeUrl
+            class="icon-block" style="font-size:1.5em; width:3em;">
+              लिंक
             </a>
+            <i class="far fa-hand-point-left"></i>
             <hr class="solid">
           </div>
         </div>
@@ -81,6 +93,9 @@ export default {
     },
     isFailSafeGform() {
       return this.value['failsafeType'] === 'g-form'
+    },
+    isVideoIdFetched() {
+      return !!this.value['youtubeId']
     }
   }
 }
--- a/src/views/Error.vue
+++ b/src/views/Error.vue
@@ -15,9 +15,8 @@
       <!-- failsafe begins -->
       <div v-if="hasFailSafe">
 
-        <!-- TODO - Adding loading spinner till YT video loads -->
         <!-- failsafe G-form begins -->
-        <div v-if="isFailSafeGform && isVideoIdFetched" >
+        <div v-if="isFailSafeGform && isVideoIdAvailable" >
           <hr class="solid">
 
           <div class="lead_text">
@@ -39,6 +38,17 @@
         <!-- failsafe G-form ends -->
 
       </div>
+
+      <div v-else-if="!hasFailSafe && isVideoIdAvailable">
+        <hr class="solid">
+
+        <div class="lead_text">
+          <p>नीचे दिया गया वीडियो देखें</p>
+          <div class='embed-container'>
+            <iframe :src="this.value['youtubeId']" id="lesson-video" frameborder='0' allowfullscreen></iframe>
+          </div>
+        </div>        
+      </div>
       <!-- failsafe ends -->
 
       <!-- TODO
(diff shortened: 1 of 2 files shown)


src/views/Error.vue
```diff
@@ -14,6 +14,9 @@
 
       <!-- TODO
       remove hardcoded link -> either parameterize it, or enter a link to youtube -->
+      <div class="lead_text">
+        <p>Plio को कैसे इस्तेमाल करना है, जानने के लिए ये video देखें </p>
+      </div>
```
deepansh96, line 19: Added text over the tutorial video embed

src/views/Error.vue
```diff
@@ -22,15 +25,24 @@
       <!-- failsafe begins -->
       <div v-if="hasFailSafe">
 
+        <!-- TODO - Adding loading spinner till YT video loads -->
```
deepansh96, line 28: As the loading up of the Youtube embed takes some time (because the youtube id is being fetched from a get request), need to add a loading spinner here as well

src/views/Error.vue
```diff
@@ -22,15 +25,24 @@
...
         <!-- failsafe G-form begins -->
-        <div v-if="isFailSafeGform">
+        <div v-if="isFailSafeGform && isVideoIdFetched" >
```
deepansh96, line 30: Added a computed property - `isVideoIdFetched`

src/views/Error.vue
```diff
@@ -22,15 +25,24 @@
...
+          <hr class="solid">
+
           <div class="lead_text">
-            <p>इसी प्रतियोगिता में Google फॉर्म से हिस्सा लेने के लिए नीचे बटन पे क्लिक करें</p>
+            <p>इसी प्रतियोगिता में Google फॉर्म से हिस्सा लेने के लिए नीचे दिया गया वीडियो देखें और उसके बाद Google फॉर्म का लिंक क्लिक करें </p>
+            <div class='embed-container'>
+              <iframe :src="this.value['youtubeId']" id="lesson-video" frameborder='0' allowfullscreen></iframe>
+            </div>
+            <br>
```
deepansh96, line 38: This is the iframe of the lesson youtube video

src/views/Error.vue
```diff
@@ -22,15 +25,24 @@
...
+            <i class="far fa-hand-point-right"></i>
             <a
-              :href=value.failsafeUrl
-              class="icon-block">
-              <img src="../assets/google_form.svg">
```
deepansh96, line 32: Removed the g-form link as we received feedback that it was non-intuitive for the students

src/views/Error.vue
```diff
@@ -22,15 +25,24 @@
...
+          <hr class="solid">
+
           <div class="lead_text">
-            <p>इसी प्रतियोगिता में Google फॉर्म से हिस्सा लेने के लिए नीचे बटन पे क्लिक करें</p>
+            <p>इसी प्रतियोगिता में Google फॉर्म से हिस्सा लेने के लिए नीचे दिया गया वीडियो देखें और उसके बाद Google फॉर्म का लिंक क्लिक करें </p>
+            <div class='embed-container'>
+              <iframe :src="this.value['youtubeId']" id="lesson-video" frameborder='0' allowfullscreen></iframe>
+            </div>
+            <br>
+            <i class="far fa-hand-point-right"></i>
```
deepansh96, line 39: added hand icons (see the screenshot in the PR description)

src/views/Error.vue
```diff
@@ -81,6 +93,9 @@ export default {
     },
     isFailSafeGform() {
       return this.value['failsafeType'] === 'g-form'
+    },
+    isVideoIdFetched() {
+      return !!this.value['youtubeId']
```
deepansh96, line 98: Will return if youtubeId has been fetched or not, and then display the `div`

src/views/Player.vue
```diff
@@ -89,7 +89,8 @@ export default {
       isBrowserSupported: true,
       browserErrorHandlingValue: {
         'failsafeType': 'g-form',
-        'failsafeUrl': ''
+        'failsafeUrl': '',
+        'youtubeId': ''
```
deepansh96, line 93: Added 'youtubeId' here because if there's a g-form, there will always be a youtube video to be seen

src/views/Player.vue
```diff
@@ -155,6 +156,7 @@ export default {
           this.videoId = res.data.videoId;
           this.plioId = res.data.plioId;
           this.browserErrorHandlingValue.failsafeUrl = res.data.plioDetails.failsafe;
+          this.browserErrorHandlingValue.youtubeId = "https://www.youtube.com/embed/" + this.videoId;
```
deepansh96, line 159: setting the URL with the fetched videoId

src/views/Error.vue
dalmia: Gave some feedback for the text

src/views/Error.vue
dalmia: Referring to the other comment, fetching probably not needed

src/views/Error.vue
dalmia: YouTube is passed as part of the value itself right? So, what is the fetch part here?

dalmia (changes requested):
Gave minor feedback on text, link view and the ordering of elements

src/views/Error.vue
deepansh96: Yes, nothing is being fetched here. Youtube ID comes from `player.vue` and it will already be present when this error page shows up. For Pritam's reference, I just used wrong terminology here. Nothing is being fetched.

dalmia: can remove this comment then right?

src/views/Error.vue
```diff
@@ -12,25 +12,27 @@
...
-        </iframe>
-      </div>
-
       <!-- failsafe begins -->
       <div v-if="hasFailSafe">
 
+        <!-- TODO - Adding loading spinner till YT video loads -->
         <!-- failsafe G-form begins -->
-        <div v-if="isFailSafeGform">
+        <div v-if="isFailSafeGform && isVideoIdFetched" >
```
dalmia, line 20: would be better to decouple video from failsafe
and rename the variable from fetch to videoIdpresent or something

dalmia (changes requested):
add minor comments

src/views/Error.vue
deepansh96: yes, will do

src/views/Error.vue
deepansh96: - Okay so that means even if there's no fail-safe g-form, the student can atleast watch the normal youtube video?
- will rename

src/views/Error.vue
pritamps: Haan nikal do

src/views/Error.vue
```diff
@@ -39,6 +41,19 @@
       </div>
       <!-- failsafe ends -->
 
+      <!-- TODO
+      remove hardcoded link -> either parameterize it, or enter a link to youtube -->
+      <div class="lead_text">
+        <p>Plio कैसे इस्तेमाल करना है, यह जानने के लिए यह video देखें </p>
+      </div>
+      <div class='embed-container'>
+        <iframe src='https://www.youtube.com/embed/FLOwzot27XM' frameborder='0' allowfullscreen>
```
pritamps, line 50: Agree that this should come as a parameter from somewhere.

Leave the TODO to change the constant, but at least put it as a property. When we want to change the URL, we wont' have to look for it deep in the code! This file is getting very big!

pritamps (approved):
This also looks good to me!! 

Just some small changes - can fix and merge!

PR #42, opened 2020-11-29
FIX: Better detection of question popup

deepansh96:
Questions that have a difference of 500ms, 400ms or even 300ms work properly now.

After this fix, I didn't notice any bugs regarding question being missed out, or popping up again.

Also added loading screen to `play` page and to individual `modals`. Reason explained in my review

Will share a demo video on Plio Tech group, cannot attach videos here

### DIFF
--- a/src/components/LoadingSpinner.vue
+++ b/src/components/LoadingSpinner.vue
@@ -1,0 +1,58 @@
+<template>
+  <div class="lds-ring">
+    <div></div>
+    <div></div>
+    <div></div>
+    <div></div>
+  </div>
+</template>
+
+<script>
+export default {
+  name: "LoadingSpinner"
+}
+</script>
+
+<style scoped>
+.lds-ring {
+  position: absolute;
+  top: 40%;
+  right: 0;
+  text-align: left;
+  display: flex;
+  justify-content: center;
+  bottom: 0;
+  left: 0;
+  z-index: 11;
+}
+.lds-ring div {
+  box-sizing: border-box;
+  display: block;
+  position: absolute;
+  width: 64px;
+  height: 64px;
+  margin: 8px;
+  border: 8px solid #fff;
+  border-radius: 50%;
+  animation: lds-ring 1.2s cubic-bezier(0.5, 0, 0.5, 1) infinite;
+  border-color: #F1C81D transparent transparent transparent;
+}
+.lds-ring div:nth-child(1) {
+  animation-delay: -0.45s;
+}
+.lds-ring div:nth-child(2) {
+  animation-delay: -0.3s;
+}
+.lds-ring div:nth-child(3) {
+  animation-delay: -0.15s;
+}
+@keyframes lds-ring {
+  0% {
+    transform: rotate(0deg);
+  }
+  100% {
+    transform: rotate(360deg);
+  }
+}
+
+</style>
--- a/src/components/PlioQuestion.vue
+++ b/src/components/PlioQuestion.vue
@@ -66,6 +66,7 @@
             v-if="isAnswerSubmitted && !isAnswerCorrect"></i>
 
           <!-- submit button -->
+          <loading-spinner v-if="showButtonLoading"></loading-spinner>
           <button
             v-if="!isAnswerSubmitted"
             class="btn submit"
@@ -77,7 +78,7 @@
 
           <!-- close button -->
           <button
-            v-else
+            v-if="!showButtonLoading && isAnswerSubmitted"
             class="btn close"
             @click="clickClose"
           >
@@ -90,7 +91,13 @@
 </template>
 
 <script>
+import LoadingSpinner from './LoadingSpinner.vue';
+
+// For how long does the spinner show (in milliseconds)
+var loadTime = 1500
+
 export default {
+  components: { LoadingSpinner },
   name: "PlioQuestion",
   props: ["plioQuestion"],
   data() {
@@ -99,14 +106,19 @@ export default {
       text: "",
       selectedOption: null,
       isAnswerCorrect: false,
-      isAnswerSubmitted: false
+      isAnswerSubmitted: false,
+      showButtonLoading: false
     };
   },
 
   computed: {
-    // Submit button disabled if no option selected
+    // Submit button disabled if no option selected or screen is loading
     isDisabled() {
-      return (this.selectedOption == null || this.isAnswerSubmitted == true);
+      return (
+        this.selectedOption == null 
+        || this.isAnswerSubmitted == true 
+        || this.showButtonLoading == true
+      );
     },
     // Returns index of the correct answer (1 indexed)
     correctAnswerIndex() {
@@ -119,6 +131,12 @@ export default {
     // Returns the index of the question that has popped up.
     currentQuestionIndex() {
       return this.plioQuestion.id;
+    },
+    reviseButton() {
+      return document.getElementById('revise-button')
+    },
+    skipButton() {
+      return document.getElementById('skip-button')
     }
   },
   methods: {
@@ -181,22 +199,30 @@ export default {
         markers[index].classList.add('tooltip-answered')
       }
     },
-
-    // Things to do after submit is clicked
+    
     clickSubmit(){
+      // Things to do after clicking submit
+      // 1-Remove old option highlights
+      // 2-Toggle the current marker from red to green
+      // 3-Show Loading Spinner and disable revise/skip/submit buttons
+      // 4-After some "loadTime", hide the revise/skip buttons
+      // 5-Remove the loading spinner, check answer and show new result
 
-      // On clicking submit, previous highlights should be removed
-      // so new highlights can be configured
       this.removeOptionHighlight();
       this.toggleMarkers(this.currentQuestionIndex)
+      this.showButtonLoading = true
+      this.reviseButton.classList.add("disabled-div")
+      this.skipButton.classList.add("disabled-div")
 
-      this.isAnswerSubmitted = true;
-
-      document.getElementById('skip-button').hidden = true
-      document.getElementById('revise-button').hidden = true
+      setTimeout(() => {
+        this.reviseButton.hidden = true
+        this.skipButton.hidden = true
+        this.isAnswerSubmitted = true;
+        this.showButtonLoading = false;
+        this.checkAnswer();
+        this.showResult();
+      }, loadTime);
       
-      this.checkAnswer();
-      this.showResult();
     },
 
     // Things to do after answer is submitted
@@ -267,6 +293,11 @@ i.fas{
   margin-top: 16px;
   margin-right: 20px;
   cursor: pointer;
+}
+
+.disabled-div {
+  pointer-events: none;
+  opacity: 0.4;
 }
 
 .leftright {
--- a/src/views/Player.vue
+++ b/src/views/Player.vue
@@ -99,7 +99,7 @@ export default {
         'failsafeUrl': ''
       },
       journey: [],
-      hasVideoPlayed: -1,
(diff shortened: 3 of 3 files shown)


src/components/LoadingSpinner.vue
```diff
@@ -0,0 +1,58 @@
+<template>
```
deepansh96, line 1: Spinner component

src/components/PlioQuestion.vue
```diff
@@ -66,6 +66,7 @@
             v-if="isAnswerSubmitted && !isAnswerCorrect"></i>
 
           <!-- submit button -->
+          <loading-spinner v-if="showButtonLoading"></loading-spinner>
```
deepansh96, line 69: Spinner will show only when `showButtonLoading` is true

src/components/PlioQuestion.vue
```diff
@@ -77,7 +78,7 @@
 
           <!-- close button -->
           <button
-            v-else
+            v-if="!showButtonLoading && isAnswerSubmitted"
```
deepansh96, line 81: `close` button (which comes after pressing `submit` button) will show only if the answer has been submitted and loading button is not showing

src/components/PlioQuestion.vue
```diff
@@ -90,7 +91,13 @@
 </template>
 
 <script>
+import LoadingSpinner from './LoadingSpinner.vue';
+
+// For how long does the spinner show (in milliseconds)
+var loadTime = 1500
```
deepansh96, line 97: This time has been set because there was a bug occurring earlier.
Earlier : User would select answer, press submit, close button will show up, press close, and modal would close. 
BUG: If the close button was pressed instantly after pressing submit, the player would break, come out of fullscreen, throw an error and would pause (these things happened inconsistently). 

Solution: Apparently we needed to wait for some time. Hence the loading spinner on the question modal.

src/components/PlioQuestion.vue
```diff
@@ -99,14 +106,19 @@ export default {
...
   computed: {
-    // Submit button disabled if no option selected
+    // Submit button disabled if no option selected or screen is loading
     isDisabled() {
-      return (this.selectedOption == null || this.isAnswerSubmitted == true);
+      return (
+        this.selectedOption == null 
+        || this.isAnswerSubmitted == true 
+        || this.showButtonLoading == true
+      );
```
deepansh96, line 121: Disable all the other buttons when there's a loading spinner running on the question pop-up

src/components/PlioQuestion.vue
```diff
@@ -119,6 +131,12 @@ export default {
     // Returns the index of the question that has popped up.
     currentQuestionIndex() {
       return this.plioQuestion.id;
+    },
+    reviseButton() {
+      return document.getElementById('revise-button')
+    },
+    skipButton() {
+      return document.getElementById('skip-button')
```
deepansh96, line 139: Added these as computed properties as the call was being repeated in the code and the browser had to go and find it again and again

src/components/PlioQuestion.vue
```diff
@@ -181,22 +199,30 @@ export default {
...
-
-    // Things to do after submit is clicked
+    
     clickSubmit(){
+      // Things to do after clicking submit
+      // 1-Remove old option highlights
+      // 2-Toggle the current marker from red to green
+      // 3-Show Loading Spinner and disable revise/skip/submit buttons
+      // 4-After some "loadTime", hide the revise/skip buttons
+      // 5-Remove the loading spinner, check answer and show new result
```
deepansh96, line 209: Comments for ease of understanding

src/components/PlioQuestion.vue
```diff
@@ -181,22 +199,30 @@ export default {
...
+        this.reviseButton.hidden = true
+        this.skipButton.hidden = true
+        this.isAnswerSubmitted = true;
+        this.showButtonLoading = false;
+        this.checkAnswer();
+        this.showResult();
+      }, loadTime);
       
-      this.checkAnswer();
-      this.showResult();
```
deepansh96, line 199: To get an idea of this, read the comment at the beginning for this method

src/components/PlioQuestion.vue
```diff
@@ -269,6 +295,11 @@ i.fas{
   cursor: pointer;
 }
 
+.disabled-div {
+  pointer-events: none;
+  opacity: 0.4;
+}
+
```
deepansh96, line 302: Added to the class list of skip/revise/submit. We want these buttons to be disabled when loading spinner is showing up

src/views/Player.vue
```diff
@@ -1,6 +1,9 @@
 <template>
-  <div> 
+  <div>
     <div class="player_container" v-if="dataLoaded && isBrowserSupported">
+
+      <LoadingSpinner v-if="!hasPlyrLoaded"></LoadingSpinner>
+
```
deepansh96, line 6: Added the spinner. Will only show if plyr has not been loaded

src/views/Player.vue
```diff
@@ -60,14 +64,17 @@ if (!Array.prototype.indexOf)
 // in milliseconds
 var interval_time = 50;
 
+// How precisely should the question pop-up logic
+// be measured. Time in milliseconds
+var popUpPrecisionTime = 10;
+
```
deepansh96, line 70: earlier we were relying on the `timeupdate` event provided by `plyr` but that only repeated every `50ms`. But it was not enough to handle those questions which were very close (< `500ms` difference).

Added another `setInterval` and the repetition time of that `setInterval` is this `popUpPrecisionTime `

src/views/Player.vue
```diff
@@ -94,7 +101,8 @@ export default {
       journey: [],
       hasVideoPlayed: -1,
       sessionId: 1,
-      hasPlyrLoaded: false
+      hasPlyrLoaded: false,
+      retention: []
```
deepansh96, line 105: Apparently a retention array was not defined as a data property, still was working fine though, no idea how.
Also there's a bug that sometimes when the response is uploaded on S3, there seems to be an extra `null` appended to the array. (Note: no where in the code am I calling `.push()` onto this array)
I'm hoping this would fix the bug.
I have not been able to replicate the bug

src/views/Player.vue
```diff
@@ -255,9 +264,6 @@ export default {
     },
 
     submitAnswer(plioQuestion, answer) {
-      // start playing whenever the user submits an answer
-      this.player.play()
-
```
deepansh96, line 260: moved this to the end.
Why? 
Because whenever the answer was submitted, the first thing that happened is the player started playing. The rest of the process was happening later. The problem with player started playing instantly was that the currentTime starts getting updated, and if another question is very very close to the current one, it again brings up those bugs.

I'm not sure how much optimization will this bring, but seems logical.

src/views/Player.vue
```diff
@@ -283,14 +289,14 @@ export default {
...
+      // start playing whenever the user submits an answer
+      this.player.play()
     },
 
     skipAnswer(plioQuestion) {
       var currQuesIndex = Number(plioQuestion.id)
 
-      // start playing if the user skips the answer
-      this.player.play()
-
```
deepansh96, line 293: Same, moved the play command to the end

src/views/Player.vue
```diff
@@ -315,8 +324,15 @@ export default {
...
-      this.player.currentTime = (currQuesIndex == 0) ? 0 : this.times[currQuesIndex - 1];
+      // after revise is clicked, make the user land just next to the marker
+      // and not on the marker so that question pops up again
+      this.player.currentTime = 
+        (currQuesIndex == 0) ? 0 : (this.times[currQuesIndex - 1] + (popUpPrecisionTime)/1000);
+
       this.player.play();
+
+      // logging for testing
+      console.log("Answer revised");
```
deepansh96, line 335: Revised the `revise` button functionality. When clicked revise, the user should not land at some previous timestamp where he will see the previous question pop-up again. We want to be just a little ahead.

src/views/Player.vue
```diff
@@ -359,19 +375,15 @@ export default {
...
           var question = plioQuestion.item;
           // Add marker to progress bar
           var marker = document.createElement("SPAN");
           marker.setAttribute("id", "marker")
           marker.classList.add("tooltip");
-          marker.classList.remove();
           var pos_percent = 100 * question.time / player.duration; 
           marker.style.setProperty("left", `${pos_percent}%`);
           progressBar.appendChild(marker);
-          //plioQuestion["marker"] = marker;
```
deepansh96, line 374: Removed unused lines

src/views/Player.vue
```diff
@@ -403,9 +415,16 @@ export default {
...
+
+          if (!instance.fullscreen.active) instance.fullscreen.enter()
+
+          if (!this.supportedBrowsers.includes(this.browser)) {
+            this.checkBrowserSupport();
+          }
+          else {
+            this.hasVideoPlayed = true
+            this.isBrowserSupported = true
+          }
```
deepansh96, line 427: Everytime play event was fired, it was trying to make it fullscreen, which gave exception sometimes.
Now it will only go to fullscreen if it's already not in fullscreen.

Also, if the current browser is supported, we'll skip the `checkBrowserSupport` check

src/views/Player.vue
```diff
@@ -426,33 +445,24 @@ export default {
...
-        if(this.player.playing) {
-          this.watchTime += interval_time;
-        }
-
-        // Record how many times a particular second was visited
-        var currTime = Math.trunc(this.player.currentTime);
-        if (currTime != prevTime) {
-            this.retention[currTime] += 1;
-            prevTime = currTime
-        }
```
deepansh96, line 444: This piece of code is showing removed here, but it's down there if you see. Github has flagged this as removed, which is not the case

src/views/Player.vue
```diff
@@ -426,33 +445,24 @@ export default {
...
-            //&& plioQuestion["state"] == "notshown"
+            // if the seeker is within "popUpPrecisionTime" of the specific question time
+            // then fire this logic
+            this.player.currentTime >= t
+            && this.player.currentTime < t + (popUpPrecisionTime)/1000
           ) {
+            this.player.pause();
             var id = plioQuestion.id
             this.$refs['position' + id.toString()].openModal();
             while (!document.querySelector(".modal")) {
```
deepansh96, line 468: if `currentTime` is within the range `[t, t+popUpPrecisionTime)`, it will open the modal.

Also moved the line `this.player.pause();` upwards because when it was down the lines, the player didn't pause for a few seconds and kept playing, and that was the reason close questions were being missed

src/views/Player.vue
```diff
@@ -464,14 +474,28 @@ export default {
 
               document
                 .getElementsByClassName("plyr")[0]
-                .appendChild(document.getElementsByClassName("modal")[0]);
+                .appendChild(modal);
```
deepansh96, line 477: `modal` variable was already defined. Hence using the variable

pritamps (approved):
This is great!

src/views/Player.vue
dalmia: Default of `isBrowserSupported` is already `True` -> So, this is not needed right?
Also, we should be consistent in what we use for `hasVideoPlayed`. In `checkBrowserSupport`, it is a binary int, here it is a `bool`.

dalmia: Default of `isBrowserSupported` is already `True` -> So, this is not needed right?
Also, we should be consistent in what we use for `hasVideoPlayed`. In `checkBrowserSupport`, it is a binary int, here it is a bool.

dalmia (changes requested):

src/views/Player.vue
deepansh96: Correct and correct. Fixing and pushing

dalmia (approved):

PR #44, opened 2020-12-05
FIX: fixed some css added some css

deepansh96:


src/components/PlioQuestion.vue
```diff
@@ -16,8 +16,8 @@
               id="skip-button"
               @click="clickSkip"
             >
-              <div class="leftright"></div>
-              <div class="rightleft"></div>
+              <i class="fas fa-window-close" 
+                 style="font-size: 1.3em; color: crimson;"></i>
```
pritamps, line 20: Add style to CSS, not in template

src/components/PlioQuestion.vue
```diff
@@ -443,47 +451,68 @@ input {
...
     text-align: center;
     text-decoration: none;
     margin: 4px 2px;
     transition-duration: 0.4s;
     cursor: pointer;
+    height: 2.5em;
+    font-size: 1.4rem;
   }
 
   .submit {
```
pritamps, line 469: Just a note. start learning SCSS. CSS will look much cleaner!

pritamps (changes requested):

src/components/PlioQuestion.vue
deepansh96: I thought SCSS can't run in all/old/stupid browsers. 
SCSS needs to be converted into normal CSS right? And vue does that automatically or are we using some package for that?

src/components/PlioQuestion.vue
deepansh96: Whoops, sorry forgot!

src/components/PlioQuestion.vue
pritamps: Yes yes, Vue will do it automatiaclly. Or we'll need a package (that won't be included in the build -- will only run while building)

I don't think any browser can read SCSS. 

ANyway, hugo ka website ka CSS SCSS mein hi hai. Seekh hi loge :D

pritamps (approved):

PR #45, opened 2021-01-02
NEW: progress will be saved across sessions

deepansh96:
Progress will be persisted across sessions.

A user coming back to a specific plio, will get his/her progress persisted.

## What will be the same? 
- A new session id will be created (as before)
- A new JSON will be created (as before)

## What will be persisted and how?
- Answers will be persisted. For every question, only the VERY FIRST recorded answer is stored.
- This VERY FIRST recorded answer will be shown to the user (red/green highlighting) whenever they come back (as a new session) and re-visit a question. They can answer it again, but that will not change the data that's stored in the back.
- The questions that have been answered in previous sessions will be shown as green markers on the progress bar when the video loads, and the unanswered ones will stay as red.
- In a new session, the video will start playing ~ 5 seconds before where the user left the video in the previous session (like in youtube)
- Watch time will be accumulated. It keeps adding up for each session.
- Retention array will be persisted and new values will be added upon the existing array.
- journey will be persisted, and new journey steps will be added to the already existing ones.

## What will NOT be persisted?
- device/browser info, session-id, source will not be persisted as it makes sense to keep them unique in every session.

### DIFF
--- a/src/components/PlioQuestion.vue
+++ b/src/components/PlioQuestion.vue
@@ -32,6 +32,7 @@
                       name="options"
                       v-model="selectedOption"
                       :value="option"
+                      :id="option"
                       @click="selectOption(index)"
                     />{{ option }}
                   </label>
@@ -149,6 +150,10 @@ export default {
       // Wait 200 ms because it takes some time to find the DOM elements
       if (this.plioQuestion.state == "answered") {
         setTimeout(() => {
+          // highlight wrong/right depending on what the user answered in previous session
+          document.getElementById(this.plioQuestion.user_answer).checked = true
+          this.selectedOption = this.plioQuestion.user_answer
+          this.checkAnswer()
           this.showResult();
           this.selectedOption = null;
         }, 200);
--- a/src/components/PlioQuestion.vue
+++ b/src/components/PlioQuestion.vue
@@ -32,7 +32,6 @@
                       name="options"
                       v-model="selectedOption"
                       :value="option"
-                      :id="option"
                       @click="selectOption(index)"
                     />{{ option }}
                   </label>
@@ -151,7 +150,8 @@ export default {
       if (this.plioQuestion.state == "answered") {
         setTimeout(() => {
           // highlight wrong/right depending on what the user answered in previous session
-          document.getElementById(this.plioQuestion.user_answer).checked = true
+          var selectedOption = document.querySelectorAll(`input[value='${this.plioQuestion.user_answer}']`)[0];
+          selectedOption.checked = true
           this.selectedOption = this.plioQuestion.user_answer
           this.checkAnswer()
           this.showResult();
--- a/src/views/Player.vue
+++ b/src/views/Player.vue
@@ -192,7 +192,7 @@ export default {
           this.times = res.data.times
 
           // merge the previous session data
-          if (res.data.sessionData != null) {
+          if (res.data.sessionData != '') {
             this.answers = res.data.sessionData.answers
             for (i=0; i < questions.length; i++){
               this.plioQuestions[i].user_answer = this.answers[i]
--- a/src/views/Player.vue
+++ b/src/views/Player.vue
@@ -203,7 +203,7 @@ export default {
             this.journey = res.data.sessionData.journey
             this.previousPlayerTime = (
               (this.journey.length > 0) ? this.journey[this.journey.length - 1]['player_time'] : 0)
-            this.watchTime += res.data.sessionData['watch-time']
+            this.watchTime = res.data.sessionData['watch-time']
             this.retention = res.data.sessionData.retention
           }
         })
@@ -402,7 +402,7 @@ export default {
         }
 
         // start from beginning if video was watched till the end in the last session
-        if (this.previousPlayerTime == this.player.duration) {
+        if (this.player.duration - this.previousPlayerTime <= 2) {
           this.player.currentTime = 0
         }
 
--- a/src/views/Player.vue
+++ b/src/views/Player.vue
@@ -62,23 +62,27 @@ if (!Array.prototype.indexOf)
 
 // The time period in which Plyr timeupdate event repeats
 // in milliseconds
-var interval_time = 50;
+const INTERVAL_TIME = 50;
 
 // How precisely should the question pop-up logic
 // be measured. Time in milliseconds
-var popUpPrecisionTime = 500;
+const POP_UP_PRECISION_TIME = 500;
 
 // upload to s3 after a fixed interval of time 
-var upload_interval = 45000;
-var timeout = null;
+const UPLOAD_INTERVAL = 45000;
+var TIMEOUT = null;
 
 // wait this much time (secs) then show error page
 // if browser is not supported
-var browserCheckTime = 10;
+const BROWSER_CHECK_TIME = 10;
 
 // how many seconds to step back (currentTime) when a user comes back to
 // a plio for a new session
-var stepBackTime = 5;
+const STEP_BACK_TIME = 5;
+
+// This buffer time is because currentTime and duration cannot be
+// exactly equal. Hence taking a ballpark of 2 seconds
+const COMPLETED_BUFFER_TIME = 2;
 
 export default {
   name: "Player",
@@ -151,7 +155,7 @@ export default {
 
     logData() {
       if (this.plioId != undefined && this.player.playing) this.uploadJson()
-      timeout = setTimeout(this.logData, upload_interval)
+      TIMEOUT = setTimeout(this.logData, UPLOAD_INTERVAL)
     },
 
     fetchData() {
@@ -194,15 +198,20 @@ export default {
           // merge the previous session data
           if (res.data.sessionData != '') {
             this.answers = res.data.sessionData.answers
-            for (i=0; i < questions.length; i++){
-              this.plioQuestions[i].user_answer = this.answers[i]
-              this.plioQuestions[i].state = 
-                (this.answers[i].length == 0) ? "notshown" : "answered"
-            }
+
+            questions.forEach((question, index) => {
+              this.plioQuestions[index].user_answer = this.answers[index]
+              this.plioQuestions[index].state = 
+                (this.answers[index].length == 0) ? "notshown" : "answered"
+            });
             
             this.journey = res.data.sessionData.journey
-            this.previousPlayerTime = (
-              (this.journey.length > 0) ? this.journey[this.journey.length - 1]['player_time'] : 0)
+
+            this.previousPlayerTime = 0
+            if (this.journey.length > 0){
+              this.previousPlayerTime = this.journey[this.journey.length - 1]['player_time']
+            }
+            
             this.watchTime = res.data.sessionData['watch-time']
             this.retention = res.data.sessionData.retention
           }
@@ -349,7 +358,7 @@ export default {
       // after revise is clicked, make the user land just next to the marker
       // and not on the marker so that question pops up again
       this.player.currentTime = 
-        (currQuesIndex == 0) ? 0 : (this.times[currQuesIndex - 1] + (popUpPrecisionTime)/1000);
+        (currQuesIndex == 0) ? 0 : (this.times[currQuesIndex - 1] + (POP_UP_PRECISION_TIME)/1000);
 
       this.player.play();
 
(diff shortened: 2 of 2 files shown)


src/components/PlioQuestion.vue
```diff
@@ -32,6 +32,7 @@
                       name="options"
                       v-model="selectedOption"
                       :value="option"
+                      :id="option"
```
deepansh96, line 35: Added `id` just so I could reference it in the code and select that radio button. This is needed because the option that was selected by the user in an older session, needs to be shown when a question is revisited.

src/components/PlioQuestion.vue
```diff
@@ -149,6 +150,10 @@ export default {
       // Wait 200 ms because it takes some time to find the DOM elements
       if (this.plioQuestion.state == "answered") {
         setTimeout(() => {
+          // highlight wrong/right depending on what the user answered in previous session
+          document.getElementById(this.plioQuestion.user_answer).checked = true
+          this.selectedOption = this.plioQuestion.user_answer
+          this.checkAnswer()
```
deepansh96, line 156: When re-visiting a question that was already answered in an older session, these steps need to followed:
1. Check the radio button corresponding to the option which was answered in the older session
2. set that value in variable `selectedOption`
3. Call `checkAnswer` method so it can detect weather the answer was correct/not and can highlight with colors accordingly

src/views/Player.vue
```diff
@@ -103,7 +103,8 @@ export default {
       hasVideoPlayed: -1, // Three possible values: -1(don't know), 0(didn't play), 1(played)
       sessionId: 1,
       hasPlyrLoaded: false,
-      retention: []
+      retention: [],
+      previousPlayerTime: 0
```
deepansh96, line 107: This variable is needed to store the `player_time` when the user left the last session. This variable is used when the player has been initialized and it's ready to play.

src/views/Player.vue
```diff
@@ -186,6 +187,22 @@ export default {
...
+              this.plioQuestions[i].state = 
+                (this.answers[i].length == 0) ? "notshown" : "answered"
+            }
+            
+            this.journey = res.data.sessionData.journey
+            this.previousPlayerTime = (
+              (this.journey.length > 0) ? this.journey[this.journey.length - 1]['player_time'] : [])
+            this.watchTime += res.data.sessionData['watch-time']
+            this.retention = res.data.sessionData.retention
+          }
```
deepansh96, line 205: If previous session data is available,
1. copy the answers into `this.answers`
2. copy the answers into the `plioQuestions` object as this is what is passed to the `PlioQuestion.vue` component
3. Also set the state as `answered` if the question was answered by the student in previous sessions
4. copy the journey
5. take the `player_time` from the very last item from `journey` array. This is where the new session's video will start playing from
6. add the watch-time to the existing watch-time
7. copy the retention array

src/views/Player.vue
```diff
@@ -376,13 +393,20 @@ export default {
 
     async setPlayerProperties(player) {
       player.on("ready", () => {
+        this.player.currentTime = Math.abs(this.previousPlayerTime - 5)
+
```
deepansh96, line 397: Start playing from where the user left in the previous session.
TODO:
If the video is finished, or the user has reached the end, start playing from beginning

src/views/Player.vue
```diff
@@ -376,13 +393,20 @@ export default {
...
           var question = plioQuestion.item;
           // Add marker to progress bar
           var marker = document.createElement("SPAN");
           marker.setAttribute("id", "marker")
           marker.classList.add("tooltip");
+
+          if (plioQuestion.state == "answered") {
+            marker.classList.add("tooltip-answered")
+          }
+
```
deepansh96, line 409: When initialzing the markers on the progress bar, check if a question has been answered before. If it is, then set the marker as green

src/views/Player.vue
```diff
@@ -396,10 +420,9 @@ export default {
           progressBar.firstChild.disabled = true;
 
         // initializing the retention array with zeros
-        this.retention = Array(this.player.duration).fill(0);
-        this.submitted = Array(this.plioQuestions.length).fill(0);
-        this.skipped = Array(this.plioQuestions.length).fill(0);
-        this.revised = Array(this.plioQuestions.length).fill(0);
```
deepansh96, line 402: Removed these three lines as this was related to an older implementation and were redundant now.

src/views/Player.vue
```diff
@@ -396,10 +420,9 @@ export default {
...
+        if (this.retention.length == 0){
+          this.retention = Array(this.player.duration).fill(0);
+        }
```
deepansh96, line 425: If we already have a persisted retention array, don't initialize it

src/views/Player.vue
```diff
@@ -376,13 +396,31 @@ export default {
...
+          this.player.currentTime = this.previousPlayerTime - stepBackTime
+        }
+        else{
+          this.player.currentTime = 0
+        }
+
+        // start from beginning if video was watched till the end in the last session
+        if (this.previousPlayerTime == this.player.duration) {
+          this.player.currentTime = 0
+        }
```
deepansh96, line 410: Start from the beginning if the video reached the end in the previous session

src/views/Player.vue
```diff
@@ -376,13 +396,28 @@ export default {
 
     async setPlayerProperties(player) {
       player.on("ready", () => {
+        // start playing from 5 seconds before where the user left off in previous session
+        if (this.previousPlayerTime > stepBackTime){
+          this.player.currentTime = this.previousPlayerTime - stepBackTime
+        }
```
deepansh96, line 402: Writing this explicitly because even if we would've written like this
```
this.player.currentTime = Math.max((this.previousPlayerTime - stepBackTime), 0)
```
Then the issue would be that it would explicitly set `currentTime` to `0` and `plyr` understands that as seeking, so when the video starts, the ugly recommendations are shown. 

If we do it like I've written, it will not set `currentTime` to `0` explicitly and it will start normally from 0.

src/views/Player.vue
```diff
@@ -186,6 +190,22 @@ export default {
...
+              this.plioQuestions[i].state = 
+                (this.answers[i].length == 0) ? "notshown" : "answered"
+            }
+            
+            this.journey = res.data.sessionData.journey
+            this.previousPlayerTime = (
+              (this.journey.length > 0) ? this.journey[this.journey.length - 1]['player_time'] : 0)
+            this.watchTime += res.data.sessionData['watch-time']
+            this.retention = res.data.sessionData.retention
+          }
```
deepansh96, line 208: - copy the answers array
- copy the answers into `plioQuestions[x].user_answer` as well
- also set the state as `answered` or `notshown`
- copy the journey array
- set the player previous time which will be used later to set the current time
- add the watch time
- copy the retention

src/views/Player.vue
```diff
@@ -74,9 +74,12 @@ var timeout = null;
 
 // wait this much time (secs) then show error page
 // if browser is not supported
-
 var browserCheckTime = 10;
 
+// how many seconds to step back (currentTime) when a user comes back to
+// a plio for a new session
+var stepBackTime = 5;
+
```
deepansh96, line 82: This means if a user watched till time `x` in a session, and they come back to a new session, the video will start playing for them from `x-5` seconds

src/views/Player.vue
```diff
@@ -376,13 +396,28 @@ export default {
...
       player.on("ready", () => {
+        // start playing from 5 seconds before where the user left off in previous session
+        if (this.previousPlayerTime > stepBackTime){
+          this.player.currentTime = this.previousPlayerTime - stepBackTime
+        }
+
+        // start from beginning if video was watched till the end in the last session
+        if (this.player.duration - this.previousPlayerTime <= 2) {
+          this.player.currentTime = 0
+        }
```
deepansh96, line 407: Eariler the code was
```if(this.player.previousPlayerTime == this.player.duration)
        this.player.currentTime = 0```
But while testing I realised that the `currentTime` or `previousPlayerTime` will never be exactly equal to the `player.duration`. So kept 2 seconds as buffer

**Comment by pritamps** [comment]
I don't get why only first answer is recorded? Don't we want to update the user's answer as they learn more?

File: src/views/Player.vue (hunk 12 above)
Comment (pritamps) [reply src/views/Player.vue:82]: `var` nahi `let` use karo, since it's a constant

File: src/components/PlioQuestion.vue (hunk 1 above)
Comment (pritamps) [reply src/components/PlioQuestion.vue:35]: But is it different from `value`?

File: src/views/Player.vue [hunk 14]
```diff
@@ -74,9 +74,12 @@ var timeout = null;
 
 // wait this much time (secs) then show error page
 // if browser is not supported
-
 var browserCheckTime = 10;
 
+// how many seconds to step back (currentTime) when a user comes back to
+// a plio for a new session
+var stepBackTime = 5;
```
pritamps, line 81: Use `const` rather than `var` since it's a constant. Will use less memory. Same for `browserCheckTime` above

pritamps: Also use all caps for constants that aren't allowed to change, and define them in one place at top of file

src/views/Player.vue
```diff
@@ -186,6 +190,22 @@ export default {
 
           // set the global list of time values
           this.times = res.data.times
+
+          // merge the previous session data
+          if (res.data.sessionData != '') {
+            this.answers = res.data.sessionData.answers
+            for (i=0; i < questions.length; i++){
```
pritamps, line 197: I think this loop will look cleaner with a `question.forEach(question, index) { }` loop

src/views/Player.vue
```diff
@@ -186,6 +190,22 @@ export default {
...
+              this.plioQuestions[i].user_answer = this.answers[i]
+              this.plioQuestions[i].state = 
+                (this.answers[i].length == 0) ? "notshown" : "answered"
+            }
+            
+            this.journey = res.data.sessionData.journey
+            this.previousPlayerTime = (
```
pritamps, line 204: Will cleaner if this is an actual `if` rather than single-line, but up to you!

src/views/Player.vue
pritamps: Nice 👍🏾

src/views/Player.vue
pritamps: Make `2` a constant. `START_BUFFER_TIME`

pritamps (commented):
This looks good!! Very small changes, and then you'll be ready!

src/components/PlioQuestion.vue
deepansh96: No it isn't. I thought it's easier to select an element using an `id` but then I was able to do the same thing with using it's `value` so I'll remove the `id` here.

src/views/Player.vue
deepansh96: But in all of our constants we have used `var`.

Screenshot 2021-01-07 at 12 13 35 PM

Also if you see (here)[https://stackoverflow.com/questions/762011/whats-the-difference-between-using-let-and-var], `var` is better suited for global use and `let` is more suited for scoping in a block use

deepansh96:
> I don't get why only first answer is recorded? Don't we want to update the user's answer as they learn more?

This was a concious decision. We all agreed in a recent meeting that we will store their first ever response and we won't let them answer again.

I actually wanted them to answer again and again but Joanna said that based on her experience with weekly quizzes, students try to game the system just to get all answers right and then it's not about learning and more about the score.

And she said this can be an experiment in itself, ki allowing kids to re answer

pritamps:
> > I don't get why only first answer is recorded? Don't we want to update the user's answer as they learn more?
> 
> This was a concious decision. We all agreed in a recent meeting that we will store their first ever response and we won't let them answer again.
> 
> I actually wanted them to answer again and again but Joanna said that based on her experience with weekly quizzes, students try to game the system just to get all answers right and then it's not about learning and more about the score.
> 
> And she said this can be an experiment in itself, ki allowing kids to re answer

Ah okay. But this is not weekly quiz no? Plios are for learning, not sorting/scoring students. Anyway, not important for this PR, but I am making prediction: we'll do all these experiments and end up where coursera is now -- no scores for in-video questions :D

deepansh96:
> > > I don't get why only first answer is recorded? Don't we want to update the user's answer as they learn more?
> > 
> > 
> > This was a concious decision. We all agreed in a recent meeting that we will store their first ever response and we won't let them answer again.
> > I actually wanted them to answer again and again but Joanna said that based on her experience with weekly quizzes, students try to game the system just to get all answers right and then it's not about learning and more about the score.
> > And she said this can be an experiment in itself, ki allowing kids to re answer
> 
> Ah okay. But this is not weekly quiz no? Plios are for learning, not sorting/scoring students. Anyway, not important for this PR, but I am making prediction: we'll do all these experiments and end up where coursera is now -- no scores for in-video questions :D

Yes I think so too! In the end we'll end up doing something like that only. But yes when something related to this comes up in a meeting then I'll bring up this point again.

pritamps (approved):
Looks good to me! Great work getting this done quickly!

In my mind, it was a big change, but glad to be proven wrong with good coding :D

PR #46, opened 2021-01-04
FIX: changed button text to hindi

deepansh96:


src/components/PlioQuestion.vue
```diff
@@ -66,10 +66,10 @@
             :disabled="isDisabled"
             @click="clickSubmit"
           >
-            ✓ Submit
+          ✓ सबमिट करें
```
pritamps, line 69: This is fine -- but @deepansh96 Notion mein ek P2 task add kar do assigned to yourself :D for "Investigation for localization"

As we launch in more states, if we need more languages, we can't depend on hardcoding strings. Let us start exploring how much work it is,

pritamps (approved):

src/components/PlioQuestion.vue
deepansh96: Makes sense! I'll add that task

PR #52, opened 2021-01-16
Execution of experiment running in A/B testing

dalmia:


src/router/index.js
```diff
@@ -2,51 +2,53 @@ import { createRouter, createWebHashHistory } from "vue-router";
...
+        name: "Player",
+        component: Player,
+        query: {
+            src: ''
+        },
+        props: route => ({ experiment: route.query.experiment })
+
+    },
+    {
+        path: "/login/:id?/:type?",
```
pritamps, line 23: Add comment here explaining what `type` is

src/store/index.js
```diff
@@ -25,6 +31,12 @@ export default createStore({
         },
         [LOGOUT](state) {
             state.isLoggedIn = false;
+        },
+        [CONFIG](state) {
+            state.pending = true;
+        },
+        [CONFIG_SUCCESS](state) {
+            state.pending = false;
```
pritamps, line 39: Hmm.. makes sense makes sense.

Change variable to `configFetchedPending` (or something similar) so it's clear what is pending

src/views/ABTesting.vue
```diff
@@ -0,0 +1,74 @@
...
+                "?experimentId=" + this.$route.params.id +
+                "&userId=" + localStorage.phone;
+            axios
+                .get(url)
+                .then((res) => {
+                    // separately seting plio ID although it will be the
+                    // same as assignment for now as we might conduct interface
+                    // level changes where assignment won't be the same as plio ID
+                    this.assignment = res.data.assignment;
+                    this.userConfig = res.data.config;
```
pritamps, line 41: Rename this variable to `userExperimentsConfig` because (if I'm understanding correctly) this is only config for experiments no?

src/views/PhoneSignIn.vue
```diff
@@ -53,8 +53,12 @@ export default {
           phone: this.phone_input,
         })
         .then(() => {
-          if (this.$route.params.id) {
-            this.$router.push( {path: "/play/" + this.$route.params.id})
+          if (this.$route.params.id){
+            if (this.$route.params.type && this.$route.params.type == 'experiment') {
+              this.$router.push({path: "/experiment/" + this.$route.params.id})
```
pritamps, line 58: `type` is not used here? Shouldn't it be part of the URL? (I'm probably understanding the logic wrong!)

src/views/Player.vue
```diff
@@ -265,7 +271,8 @@ export default {
               'source': this.source,
               'retention': this.retention,
               'has-video-played': this.hasVideoPlayed,
-              'journey': this.journey
+              'journey': this.journey,
+              'experiment': this.experiment
```
pritamps, line 275: Cool!

pritamps (changes requested):
Very cool! Looking forward to experiments data :D

src/views/PhoneSignIn.vue
dalmia: The `type` here is just for the phone sign in component to know what type of component invoked it. Here, the `type == 'experiment'` check ensures that if the sign in component was invoked from the ABTesting component, it should redirect back to it. Does that make sense?

pritamps: Ah understood yes yes cool cool

src/views/ABTesting.vue
dalmia: For now, this is retrieving the whole config (it will become clearer in the backend ka PR). The point was that the user should only need to fetch the config once. IF they are coming through the experiment link, then we fetch the config here and store it in Vuex. Now, when Deepansh's code wants to check for the tutorial progress, he will first check in 
Vuex if there is a config in Vuex. If it is present, then it simply uses it. Does that make sense? I thought we'll make the segregation amongst configs when we feel the need to.

pritamps: Haan understood. Feels a little strange... Can you just add a comment in the `Login` component and router that individual components are responsible for getting the user's config if they need it.

pritamps (approved):
Looks good. Only one comment to be added ( to log in component saying other people have to pull the config) and then you can ship!

deepansh96 (approved):

src/views/ABTesting.vue
pritamps: I gave Deepansh a suggestion on the phone. Make a separate component to handle user config stuff. Then any other component can say `UserConfigComponent.fetchConfig("experiments")` and the Config component can handle fetching and all that depending on the state.

It's better to do it now than refactor later. But if some deadline

deepansh96: @dalmia what do you think?

deepansh96 (approved):

src/views/ABTesting.vue
dalmia: UserConfig please thoda alag se hi karte hai please

PR #53, opened 2021-01-16
NEW: interactive tutorial

deepansh96:
Created a bunch of components (4) for the tutorial. More can be added if required. These components are font-awesome svgs. Each one can be imported where it needs to be shown.

These components (tutorial steps) can be toggled using `v-if` conditions.

While making the initial call of getting the plio details, the config data is also passed to the frontend. And as the user goes on with the tutorial, the updated config is uploaded.

The structure of tutorial config that is being pulled/pushed is this
```
'configs' : {
        'tutorial' : {
            'isComplete' : true/false,
            'progress' : {
                  'start' : true/false,
                  'options' : true/false,
                  'submit' : true/false,
                  'close' : true/false
            }
       }
}
```

In this PR, i'm fetching the config information along with all the plio details, but after Aman's code is set and tested, will change that to pick up config information from vuex


src/components/PlioQuestion.vue
```diff
@@ -43,8 +43,11 @@
             </ul>
 
             <!-- Selected: {{ selectedOption }} -->
+            <mcqOptionsPointer 
+              v-if="!isAnAnsweredQuestion && !isTutorialComplete && !tutorialProgress['options']">
+            </mcqOptionsPointer>
```
deepansh96, line 48: This is the component that is shown pointing towards the options.
It will show only when -
(
the question has not been answered before +
the tutorial is not complete +
this exact tutorial step has not been seen before
)

src/components/PlioQuestion.vue
```diff
@@ -73,11 +76,17 @@
           >
           ✓ सबमिट करें
           </button>
+          <submit-button-pointer 
+            v-if="!isTutorialComplete && !tutorialProgress['submit'] && !isDisabled && !isAnAnsweredQuestion">
+          </submit-button-pointer>
```
deepansh96, line 81: The submit button pointer should appear only when 
(
the tutorial is not complete +
this particular step has not been seen before +
the submit button is not disabled +
this question has not been answered before
)

src/components/PlioQuestion.vue
```diff
@@ -99,13 +108,17 @@ import { faCheckCircle } from '@fortawesome/free-solid-svg-icons/faCheckCircle';
 import { faTimesCircle } from '@fortawesome/free-solid-svg-icons/faTimesCircle';
 library.add(faWindowClose, faCheckCircle, faTimesCircle);
 
+import SubmitButtonPointer from './tutorial/SubmitButtonPointer.vue';
+import mcqOptionsPointer from './tutorial/mcqOptionsPointer.vue';
+import CloseButtonPointer from './tutorial/CloseButtonPointer.vue';
```
deepansh96, line 113: Importing these tutorial components

src/components/PlioQuestion.vue
```diff
@@ -99,13 +108,17 @@ import { faCheckCircle } from '@fortawesome/free-solid-svg-icons/faCheckCircle';
...
+
 // For how long does the spinner show (in milliseconds)
 var loadTime = 1500;
 
 export default {
-  components: { LoadingSpinner },
+  components: { LoadingSpinner, SubmitButtonPointer, mcqOptionsPointer, CloseButtonPointer},
   name: "PlioQuestion",
-  props: ["plioQuestion"],
+  props: ["plioQuestion", "isTutorialComplete", "tutorialProgress"],
```
deepansh96, line 121: The information if the tutorial is complete and what the progress is, that is being passed as props to `PlioQuestion.vue`

src/components/PlioQuestion.vue
```diff
@@ -137,6 +150,9 @@ export default {
     // Returns the index of the question that has popped up.
     currentQuestionIndex() {
       return this.plioQuestion.id;
+    },
+    isAnAnsweredQuestion(){
+      return this.plioQuestion.state == "answered"
```
deepansh96, line 155: A new method to tell if this question has already been answered by the student or not

src/components/PlioQuestion.vue
```diff
@@ -160,6 +176,7 @@ export default {
       if (this.plioQuestion.state == "answered") {
         setTimeout(() => {
           // highlight wrong/right depending on what the user answered in previous session
+          document.getElementById("options_container").classList.add("options-block");
```
deepansh96, line 179: This is done because earlier the `display` of the options container was `block` but then when the tutorial message was added (`choose only one`), the `display` had to be changed to `flex` to incorporate that message as well.
BUT, when the tutorial message is not required, the `display` needs to be switched back to `block` otherwise the css breaks and it looks weird

src/components/PlioQuestion.vue
```diff
@@ -247,14 +265,16 @@ export default {
         this.checkAnswer();
         this.showResult();
       }, loadTime);
+
+      this.$emit("answer-submitted")
```
deepansh96, line 269: Earlier, nothing was being emitted when `submit` was clicked. The response was being emitted when `close` was clicked. Why change it now?
Because we want to know if the user has clicked submit -> and then mark that part of tutorial complete.
similarly, we want to know if the user has clicked close   -> and then mark that part of tutorial complete

src/components/PlioQuestion.vue
```diff
@@ -247,14 +265,16 @@ export default {
...
     },
 
     // Things to do after answer is submitted
     // and close button appears
     clickClose() {
       this.closeModal();
       this.isAnswerSubmitted = false;
-      this.$emit("answer-submitted", this.plioQuestion, this.selectedOption);
+      this.$emit("question-closed", this.plioQuestion, this.selectedOption);
```
deepansh96, line 277: As explained in the comment above, now we know when submit was clicked and close was clicked

src/views/Player.vue
```diff
@@ -153,11 +164,44 @@ export default {
       return dateTime
     },
 
+    waitFor(conditionFunction) {
+      const poll = resolve => {
+        if(conditionFunction()) resolve();
+        else setTimeout(() => poll(resolve), 400);
+      }
+      return new Promise(poll);
+    },
```
deepansh96, line 173: This is a promise based utility function that can be reused. You can give it a condititon to check and it will wait for 400ms until that condition is met. I think we can change the 400ms to a globally defined variable

src/views/Player.vue
```diff
@@ -153,11 +164,44 @@ export default {
...
+      x.classList.add('start-button-active')
+      setTimeout(() => {
+        this.player.fullscreen.enter();
+        this.waitFor(() => this.player.fullscreen.active === true)
+        .then(() => this.player.play())
+      }, 400);
+
+      this.tutorialProgress['start'] = true;
+      this.checkIfTutorialIsComplete();
+    },
```
deepansh96, line 192: When start is clicked,
- change the class of the button (so that css changes and we can see the button being "clicked")
- enter fullscreen
- wait till plyr recognizes that fullscreen is active (this was important as otherwise it was giving an exception)
- auto play
- mark the start button pointer of the tutorial as true
- check if the tutorial is complete

src/views/Player.vue
```diff
@@ -153,11 +164,44 @@ export default {
...
+    checkIfTutorialIsComplete(){
+      this.isTutorialComplete = true;
+      for (const [key, value] of Object.entries(this.tutorialProgress)) {
+        console.log(key, value);
+        if (value == false) {
+          this.isTutorialComplete = false;
+          return
+        }
+      }
+    },
```
deepansh96, line 203: iterate through the dictionary,
if all values are true, then the tutorial is complete and vice versa

src/views/Player.vue
```diff
@@ -177,6 +221,8 @@ export default {
           this.isFullscreen = false;
           this.sessionId = res.data.sessionId;
           this.browser = res.data.userAgent['browser']['family'];
+          this.isTutorialComplete = res.data.configData.tutorial.isComplete;
+          this.tutorialProgress = res.data.configData.tutorial.progress;
```
deepansh96, line 225: this is being fetched from backend

src/views/Player.vue
```diff
@@ -279,6 +325,33 @@ export default {
...
+                headers: {
+                  'Accept': 'application/json',
+                  'Content-Type': 'application/json'
+                }})
+                .then(response => response.json())
+                .then(data => console.log(data))
+                .catch(err => console.log(err))
+          
+          if (this.isTutorialComplete) this.isTutorialUploadRequired = false;
+      }
```
deepansh96, line 354: Uploading the updated config (with updated tutorial data)

src/views/Player.vue
```diff
@@ -292,12 +365,21 @@ export default {
             'system_time': String(this.getCurrentDateTime()),
             'player_time': String(player_time)
         })
+
+        if (logEvent == 'option-selected'){
+          this.tutorialProgress['options'] = true;
+          this.checkIfTutorialIsComplete();
+        }
```
deepansh96, line 372: if option is selected, mark that part of the tutorial as true

src/components/PlioQuestion.vue
pritamps: Nice :) 

The hand shows in the center of the screen. Maybe we can update (not required for thiS PR) to calculate it's position based on the button's position

src/components/PlioQuestion.vue
pritamps: Why the other conditions. Why not only the two tutorial conditions?

src/components/PlioQuestion.vue
pritamps: Each pointer is separate component? Can you make a pointer component, and keep the icon, and position as parameter?

src/components/PlioQuestion.vue
pritamps: I feel like since it's a one line method, might as well just be an inline condition (Opinion only). Will avoid the overhead of function calls. It's negligible, but still ;-)

src/components/PlioQuestion.vue
pritamps: oh boy. I will take your word for it. But usually, with CSS, there's a way around hacks. Maybe if we use `grid`? We can look later!

src/components/PlioQuestion.vue
pritamps: Cool 👍🏾

src/components/tutorial/StartButtonPointer.vue
```diff
@@ -0,0 +1,30 @@
+<template>
```
pritamps, line 1: Yeah all these template files are mostly the same exactly the same -- can you combine and parametrize? 

If not possible before tomorrow ,let it be. When we open source it, we can do it 👍🏾

src/views/Player.vue
```diff
@@ -153,11 +165,46 @@ export default {
...
+        .then(() => this.player.play())
+      }, 400);
+
+      this.tutorialProgress['start'] = true;
+    },
+
+    checkIfTutorialIsComplete(){
+      this.isTutorialComplete = true;
+      for (const [key, value] of Object.entries(this.tutorialProgress)) {
+        console.log(key, value);
```
pritamps, line 200: Remove `log` statements

src/views/Player.vue
```diff
@@ -153,11 +165,46 @@ export default {
...
+        if (value == false) {
```
pritamps, line 201: I think you can write this all in one line: 

```js
return (false in this.tutorialProgress)
``` 
 instead of the loop.

src/views/Player.vue
pritamps: I feel like this condition `isTutorialUploadRequired` should not be needed. If the tutorial is complete, none of this code should be reached because some `if` higher in the chain will block tutorial related actions. But for future

pritamps (approved):
Since this is urgent, None of the changes are actually required, except maybe removing log statements. All for consideration only!

But let's keep a link to this review somewhere so that we can refer back to it

PR #59, opened 2021-02-05
Progress bar

deepansh96:
- Player fetches the config and loads the system default config
- Cross verifies both of them and picks up the details of a key if they are present in the system default config
- initialises the variables and objects that need to be passed to `PlioQuestion.vue`
- If past session data is available, update the objects defined in the previous step
- send it to `PlioQuestion.vue`
- `PlioQuestion.vue` will use the info to tell `ProgressBar.vue` where to start when rendering
- Then it will wait for the users response, if it user responds, `PlioQuestion.vue` will update the completion % and tell `ProgressBar.vue` to also animate and increase the progress bar to a desired value
- And in the end, the updated progress bar value goes back to `Player.vue` so it can be used in the upcoming question again


src/assets/component-properties.json
```diff
@@ -0,0 +1,5 @@
+{
+  "player": {
+    
+  }
+}
```
deepansh96, line 5: The locally stored global default file. If any feature details exist here, this version of details will be used instead of the one present in the incoming plio config

src/components/PlioQuestion.vue
```diff
@@ -95,6 +95,13 @@
           आगे बढ़ें
           </button>
         </div>
+        <progress-bar 
+          v-if="isProgressBarEnabled"
+          ref="progressBarRef"
+          :startPercent="progressBarInfo['completionPercent']"
+          :config="progressBarInfo['config']"
+        >
+        </progress-bar>
```
deepansh96, line 104: Progress bar component.
- Will show only when it's enabled. (this `enabled` info comes from the `Player.vue` which fetches the plio config for this specific plio)
- `startPercent` -> this tells the progress bar to start at `x %` whenever it is rendered. 
- `config` -> this is the specific progress bar details that have been fetched by the backend

src/components/PlioQuestion.vue
```diff
@@ -127,6 +147,11 @@ export default {
       isAnswerCorrect: false,
       isAnswerSubmitted: false,
       showButtonLoading: false,
+      updatedProgressBarInfo: {
+        "config": this.progressBarInfo['config'],
+        "updatedCompletionPercent": 0,
+        "totalQuestions": this.progressBarInfo['totalQuestions']
+      }
```
deepansh96, line 154: Everytime a question interface is injected into the main HTML, this object is created for that interface. It contains the config of the progress bar, total questions in that plio and the `updatedCompletionPercent`. This `updatedCompletionPercent` is initially set to what the completion % was before this question appeared.
`PlioQuestions.vue` receives this information as props from `Player.vue`

src/components/PlioQuestion.vue
```diff
@@ -153,6 +178,9 @@ export default {
     },
     isAnAnsweredQuestion(){
       return this.plioQuestion.state == "answered"
+    },
+    isProgressBarEnabled(){
+      return this.progressBarInfo['config']['enabled']
```
deepansh96, line 183: computed property -- it takes this information from the incoming config

src/components/PlioQuestion.vue
```diff
@@ -241,6 +271,19 @@ export default {
...
+          'updatedCompletionPercent'] += ((1/this.updatedProgressBarInfo['totalQuestions'])*100)
+      }
+
+      setTimeout(() => {
+        this.$refs["progressBarRef"].progressTo(
+          this.updatedProgressBarInfo['updatedCompletionPercent']
+        );
+      }, 200)
+    },
+
```
deepansh96, line 286: 2 steps
- calculate the new progress, set it to the `updatedCompletionPercent` key of `updatedProgressBarInfo` object
- wait for 200ms, call the progress bar component's method `progressTo` which basically tells the component to progress to some value. This value is the `updatedCompletionPercent` as calculated in above point

src/components/PlioQuestion.vue
```diff
@@ -264,6 +307,7 @@ export default {
         this.showButtonLoading = false;
         this.checkAnswer();
         this.showResult();
+        this.updateAndShowProgress();
```
deepansh96, line 310: Update and show progress when question is submitted

src/components/PlioQuestion.vue
```diff
@@ -274,7 +318,8 @@ export default {
     clickClose() {
       this.closeModal();
       this.isAnswerSubmitted = false;
-      this.$emit("question-closed", this.plioQuestion, this.selectedOption);
+      this.$emit("question-closed", this.plioQuestion, 
+        this.selectedOption, this.updatedProgressBarInfo);
```
deepansh96, line 322: When `close` button is clicked, along with all other info, we're also passing the info that the progress bar has moved to a new % value so it could be used again in next question

src/components/features/ProgressBar.vue
```diff
@@ -0,0 +1,80 @@
+<template>
+  <div class="progress progress-striped">
```
deepansh96, line 2: the outer shell

src/components/features/ProgressBar.vue
```diff
@@ -0,0 +1,80 @@
...
+    <div class="progress__bar"></div>
```
deepansh96, line 3: the actual progress rectangle filling up

src/components/features/ProgressBar.vue
```diff
@@ -0,0 +1,80 @@
...
+
+  methods: {
+    initializeProgressBar(){
+      setTimeout(() => {
+        this.progressBarContainer = document.querySelector(".progress")
+        this.progressBar = document.querySelector(".progress__bar")
+        this.setConfigProperties(this.config['details'])
+        this.progressTo(this.startPercent)
+      }, 200)
+    },
```
deepansh96, line 33: select the progress bar container and the progress bar, set their config properties and set the initial position where the progress bar should be

src/components/features/ProgressBar.vue
```diff
@@ -0,0 +1,80 @@
...
+    setConfigProperties(config){
+      // for the container
+      Object.assign(this.progressBarContainer.style, {
+        background: `${config['empty_color']}`
+      })
+
+      // for the bar
+      Object.assign(this.progressBar.style, {
+        "background-color": `${config['progress_color']}`
+      })
```
deepansh96, line 44: assigning CSS according to the config passed. Right now it's just color, but if needed, more can be added

src/views/Player.vue
```diff
@@ -15,6 +15,7 @@
           :ref="'position' + plioQuestion.id.toString()"
           :isTutorialComplete="isTutorialComplete"
           :tutorialProgress="tutorialProgress"
+          :progressBarInfo="progressBarInfo"
```
deepansh96, line 18: sending the progress bar info to `PlioQuestions.vue`

src/views/Player.vue
```diff
@@ -125,7 +126,7 @@ export default {
       },
       journey: [],
       hasVideoPlayed: -1, // Three possible values: -1(don't know), 0(didn't play), 1(played)
-      sessionId: 1,
+      sessionId: 0,
```
deepansh96, line 129: missed this in earlier PR. We had a mutual decision to start sessions from 0. Changing here doesn't affect the functionality. This is just for consistency

src/views/Player.vue
```diff
@@ -134,6 +135,12 @@ export default {
       tutorialProgress: {},
       isTutorialUploadRequired: false,
       isModalOnScreen: false,
+      componentProperties: {},
```
deepansh96, line 138: a data property to store the `component-properties.json`, i.e. the system wide properties we (might) have when we decide to keep a feature always

src/views/Player.vue
```diff
@@ -134,6 +135,12 @@ export default {
...
+      progressBarInfo: {
+        "config": {},
+        "completionPercent": 0,
+        "totalQuestions": 0
+      }
```
deepansh96, line 143: this is what will be passed to the `PlioQuestions.vue`

src/views/Player.vue
```diff
@@ -226,9 +237,22 @@ export default {
           this.isFullscreen = false;
           this.sessionId = res.data.sessionId;
           this.browser = res.data.userAgent["browser"]["family"];
-          this.configs = res.data.configData;
-          this.isTutorialComplete = this.configs.tutorial.isComplete;
-          this.tutorialProgress = this.configs.tutorial.progress;
+          this.userConfigs = res.data.userConfig;
+          this.isTutorialComplete = this.userConfigs.tutorial.isComplete;
+          this.tutorialProgress = this.userConfigs.tutorial.progress;
```
deepansh96, line 242: Changed `configData` to `userConfig` because now we also have `plioConfig` so `plioConfig` would've been hard to differentiate from `configData`

src/views/Player.vue
```diff
@@ -226,9 +237,22 @@ export default {
...
+            this.plioPlayerConfig = res.data.plioConfig['player']
+          }
+          else this.plioPlayerConfig = {}
+
+          for (const [feature, details] of Object.entries(this.componentProperties)) {
+              this.plioPlayerConfig[feature] = details
+          }
+
+          this.progressBarInfo['config'] = this.plioPlayerConfig['progress_bar']
+          this.progressBarInfo['completionPercent'] = 0
```
deepansh96, line 255: 1. Storing the player specific config data into a variable
2. Iterating through the system wide config, and if anything is found, replace the info of that key into the `plioPlayerConfig`
3. set the `progress_bar` info to another object named `progressBarInfo` which will be passed on to the `PlioQuestion.vue`
4. Set the `completionPercent` as 0 as a default

src/views/Player.vue
```diff
@@ -255,6 +281,10 @@ export default {
               this.plioQuestions[index].user_answer = this.answers[index];
               this.plioQuestions[index].state =
                 this.answers[index].length == 0 ? "notshown" : "answered";
+
+              if (this.plioQuestions[index].state == "answered") {
+                this.progressBarInfo['completionPercent'] += 1
+              }
```
deepansh96, line 287: iterate through the answers of the previous session, if any question has been answered, increase the progress counter

src/views/Player.vue
```diff
@@ -268,6 +298,11 @@ export default {
 
             this.watchTime = res.data.sessionData["watch-time"];
             this.retention = res.data.sessionData.retention;
+            if (questions.length > 0) {
+              var currCompletion = this.progressBarInfo['completionPercent']
+              this.progressBarInfo['completionPercent'] = Math.ceil(
+                (currCompletion/questions.length)*100)
+            }
```
deepansh96, line 305: store the progress counter as a % to be consumed by the progress bar component

src/views/Player.vue
```diff
@@ -384,11 +419,14 @@ export default {
       }
     },
 
-    recordAnswer(plioQuestion, answer) {
+    recordAnswer(plioQuestion, answer, updatedProgressBarInfo) {
```
deepansh96, line 422: this extra info `updatedProgressBarInfo` was emitted by `PlioQuestion.vue`

src/views/Player.vue
```diff
@@ -384,11 +419,14 @@ export default {
...
       // this function is called when the close button is clicked
       // Update state to "answered"
       plioQuestion["state"] = "answered";
 
+      this.progressBarInfo['completionPercent'] = updatedProgressBarInfo[
+        'updatedCompletionPercent']
+
```
deepansh96, line 429: we're updating our current progress with the value that we received from `PlioQuestion.vue` after user has submitted and closed the question

src/views/Player.vue
```diff
@@ -500,8 +538,8 @@ export default {
           } else this.hasVideoPlayed = 1;
           this.uploadJson();
 
-          var progressBar = document.querySelectorAll(".plyr__progress")[0];
-          progressBar.firstChild.removeAttribute("disabled");
+          var plyrProgressBar = document.querySelectorAll(".plyr__progress")[0];
+          plyrProgressBar.firstChild.removeAttribute("disabled");
```
deepansh96, line 542: renaming to `plyrProgressBar` to avoid confusion

src/components/PlioQuestion.vue
pritamps: 1. This is specifically a question progress bar no? 
2. This config doesn't have default values in the default config?
3. Naming mismatch - `startPercent` vs `completionPercent`. Maybe rename both to `progressPercent`?

src/components/PlioQuestion.vue
```diff
@@ -127,6 +147,11 @@ export default {
       isAnswerCorrect: false,
       isAnswerSubmitted: false,
       showButtonLoading: false,
+      updatedProgressBarInfo: {
+        "config": this.progressBarInfo['config'],
+        "updatedCompletionPercent": this.progressBarInfo['completionPercent'],
```
pritamps, line 152: Instead of `updated` -- rename to `newCompletionPercent`. This is for the animation from old position to new position no?

src/components/PlioQuestion.vue
pritamps: Why wait 200 ms here?

src/components/PlioQuestion.vue
pritamps: Oh no... Just realized. The progress bar calculations are being done inside `PlioQuestion`. This IMO, is wrong, because it assumes that questions know about other questions (`totalQuestions` is the biggest sinner).

The correct design here is to have the progress progress be set from the player:

1. The question doesn't know about other questions
2. The player tells the progress bar how much to do, and the modal just shows it. It's actually okay for the progress bar to know this (though that means it will not be reusable in other places)

If you want, you can skip this today for merging in case it's a lot of work. But this is VERY IMPORTANT (my opinion, of course), for separation of responsibility. You can merge, test, and then make this change immediately as your next task after the A/B testing release.

src/components/features/ProgressBar.vue
pritamps: Hmm... this has to be done in code, no? :( I'm sure there's a way to do it with separation, but I don't know :D

src/components/features/ProgressBar.vue
```diff
@@ -0,0 +1,80 @@
...
+    initializeProgressBar(){
+      setTimeout(() => {
+        this.progressBarContainer = document.querySelector(".progress")
+        this.progressBar = document.querySelector(".progress__bar")
+        this.setConfigProperties(this.config['details'])
+        this.progressTo(this.startPercent)
+      }, 200)
+    },
+
+    setConfigProperties(config){
```
pritamps, line 35: Rename to `setCSSProperties`, or if you expect that other configs will be set, create a method called `setCSSProperties` and call it from here. Separating CSS parts from code parts is super important.

You should also add a comment in the CSS itself, saying that these properties are set in the code

src/views/Player.vue
pritamps: As above, don't send to `PlioQuestion`. I don't know how complicated it is, though because I can't think reading through the PR how it will work.... animation should be handled by the progress bar and player. Maybe the right answer is to have a meta-component for the modal, which will have two (or more in future) components: `PlioQuestion`, `ProgressBar`, and this meta-component can deal with both things.

src/views/Player.vue
pritamps: This needs defaults that will be populated from the `defaults` json file

src/views/Player.vue
pritamps: 👍🏾

pritamps (changes requested):

src/assets/component-properties.json
deepansh96: Default will have those details which we want to be persistent across all the plios/components. Till now we have not decided what will be visible in all the plios. 
Let's say we decide we want to finalize the progress bar with some specific properties, so we will put that information here and it will be applied to all the plios

src/components/PlioQuestion.vue
deepansh96: If I try to find the `this.$refs["progressBarRef"]` instantly without waiting, it has not been injected into the HTML until now (my guess) and it returns `null` but when I wait for some time, it successfully returns the element

deepansh96:
@pritamps I agree with you that the progress bar related calculation should be handled by `Player.vue` and the info can be passed to the progress bar and `PlioQuestion.vue` doesn't need to come into the picture. Will try to do that but won't be able to do that now. 
I started with the current design because I assumed that the progress bar will be a part of the question. If it will be called by the player, how should it be rendered in the HTML? Separately from the plioQuestion component?

PR #62, opened 2021-02-27
NEW: CMS support

deepansh96:
Frontend now will be able to render the questions received from CMS (cms.peerlearning.com)

Frontend will receive the data in the plio JSON as usual. However, the JSON received can also contain the question/option text in the form of LaTex, Kruti font, and images (in the form of an `` tag. 

These details already come pre-filled in the plio JSON. A separate `meta_data` is also added to the plio JSON but the frontend is not aware of it and doesn't make use of that.

Steps of the process are outlined here. For more granular details, please see the comments left on the code.
- plio JSON with all the details is fetched
- The question + option data is passed to `PlioQuestion.vue` where it's saved as a data property
- When the DOM is mounted, the MathJax scripts are added to the document head
- As the modal opens, a series of steps are carried out
- Incoming image, if any, is handled
- Question and option styling is applied
- The LaTex is force rendered

For more details on the whole thing: [notion doc](https://www.notion.so/avantifellows/Render-questions-using-the-CMS-or-any-other-service-4aa25eb31846408e9f2d0173e67386f4)


src/App.vue
```diff
@@ -57,4 +57,10 @@ export default {
 #nav a.router-link-exact-active {
   color: #42b983;
 }
+
+@font-face {
+  font-family: "Kruti Dev";
+  src: local("Kruti Dev"),
+    url('./fonts/Kruti_Dev_10.TTF') format("truetype");
```
deepansh96, line 64: Installing the Kruti font. Used by CMS for typing in hindi.

src/components/PlioQuestion.vue
```diff
@@ -7,8 +7,7 @@
         <div class="modal__body">
           <!-- The cross button to close the modal -->
           <div class="question_text_row">
-            <div class="question_text">
-              {{ plioQuestion.item.question.text }}
+            <div class="question_text" id="question" v-html="questionText">
```
deepansh96, line 10: Instead of rendering the text as just a string (`{{ plioQuestion.item.question.text }}`), using a `v-html` directive will tell Vue to treat the text as HTML. 
The question information coming from the CMS is actually HTML content

src/components/PlioQuestion.vue
```diff
@@ -29,14 +28,16 @@
...
                     <input
                       type="radio"
                       name="options"
                       v-model="selectedOption"
                       :value="option"
+                      :id = "'option_input_' + index"
                       @click="selectOption(index)"
-                    />{{ option }}
+                    />
+                      <div class="option_render" :id= "option" v-html="optionText[index]"></div>
```
deepansh96, line 40: As explained for the `question` part, similarly, for the `option`, the text will be rendered as a HTML

src/components/PlioQuestion.vue
```diff
@@ -151,7 +152,9 @@ export default {
         "config": this.progressBarInfo['config'],
         "progressPercent": 0,
         "totalQuestions": this.progressBarInfo['totalQuestions']
-      }
+      },
+      questionText: this.plioQuestion.item.question.text,
+      optionText: this.plioQuestion.item.question.options,
```
deepansh96, line 157: Saving the `questions` and `options` information as data properties so they can be referenced in the `v-html` directive as explained above

src/components/PlioQuestion.vue
```diff
@@ -184,6 +187,12 @@ export default {
         return this.progressBarInfo['config']['enabled']
       
       return false;
+    },
+    userAnswerIndex() {
+      // index of user answer in the options array
+      return this.plioQuestion.item.question.options.indexOf(
+        this.plioQuestion.user_answer
```
deepansh96, line 194: Returns the index of the user answer in the options array. Why this is needed now?
Earlier, when coming back to an already answered question, there was a need to select the option which was the user's answer. And that was being done by a selector which selected the element based on the option text.

But now, as the option text can contain random latex character and random strings, it's best to select those options using their index.

src/components/PlioQuestion.vue
```diff
@@ -205,10 +214,10 @@ export default {
       // Show highlighted options if coming back to an answered question
       // Wait 200 ms because it takes some time to find the DOM elements
       if (this.plioQuestion.state == "answered") {
-        setTimeout(() => {
+        requestAnimationFrame(() =>{
```
deepansh96, line 217: This eliminates the need for keeping a 200ms `setTimeOut`. More details [here](https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame)

Although this is used to handle animations, it works as a hack in a lot of cases where we need to repaint the screen and we ask the browser to finish all its tasks and go into this block only when everything is updated/painted on the screen.

src/components/PlioQuestion.vue
```diff
@@ -205,10 +214,10 @@ export default {
...
           // highlight wrong/right depending on what the user answered in previous session
           document.getElementById("options_container").classList.add("options-block");
-          var selectedOption = document.querySelectorAll(`input[value='${this.plioQuestion.user_answer}']`)[0];
+          var selectedOption = document.getElementById(`option_input_${this.userAnswerIndex}`)
```
deepansh96, line 220: Old selector : option text
New selector : user's answer index

src/components/PlioQuestion.vue
```diff
@@ -218,13 +227,88 @@ export default {
...
       document.querySelector("body").classList.add("overflow-hidden");
       this.newProgressBarInfo['progressPercent'] = this.progressBarInfo['progressPercent']
+      this.handleImage()
+      this.styleQuestion()
+      this.styleOptions()
+      this.renderContent()
+    },
+
+    handleImage(){
+      this.$nextTick(() => {
```
deepansh96, line 244: This method executes the code only after the DOM updation cycle has finished. Will only execute the code when everything in the DOM is ready and updated so we won't face any issues while running `document.getElementById` type of searches, hence, removing the need of a `setTimeOut` method

src/components/PlioQuestion.vue
```diff
@@ -218,13 +227,88 @@ export default {
...
+        var p_tags = question.querySelectorAll('p')
+        if (p_tags == null || p_tags.length == 0) return
+        // find which <p> tag has an <img> tag, cut it and
+        // paste it right outside, just below it.
+        p_tags.forEach(p_tag => {
+          var img_tag = p_tag.querySelector('img')
+          if (img_tag != null) {
+            p_tag.removeChild(img_tag)
+            p_tag.parentNode.insertBefore(img_tag, p_tag.nextSibling)
+            img_tag.setAttribute("id", "question_image")
```
deepansh96, line 258: Removes the `img` from the `p` tag and places it just outside, below the same `p` tag.
Note: in JS, there's no `insertAfter` method. Hence there is a workaround using the `insertBefore` method. More details [here](https://www.javascripttutorial.net/javascript-dom/javascript-insertafter/)

src/components/PlioQuestion.vue
```diff
@@ -218,13 +227,88 @@ export default {
...
+          }
+        })
+      })
+    },
+
+    renderContent(){
+      // force rendering of question and option texts
+      this.$nextTick(() => {
+        MathJax.Hub.Queue(["Typeset",MathJax.Hub,"question"]);
+        MathJax.Hub.Queue(["Typeset",MathJax.Hub,"option"]);
```
deepansh96, line 273: Here, `question` and `option` are the `id` attributes of the tags that need to be rendered using MathJax

src/components/PlioQuestion.vue
```diff
@@ -218,13 +227,88 @@ export default {
...
+        if (question == null) return
+  
+        var p_tags = question.querySelectorAll('p')
+        if (p_tags == null || p_tags.length == 0) return
+
+        p_tags.forEach(p_tag => {
+          p_tag.style.lineHeight = '100%'
+          p_tag.style.marginRight = 'auto'
+          p_tag.style.marginBlock = '5px';
+        })
```
deepansh96, line 291: The styling here is done in JS because the `<p>` tags by default have a styling (provided by the browser). It was not possible to override the default styles in CSS because the `<p>` tags are created only afterwards.

src/components/PlioQuestion.vue
```diff
@@ -279,11 +363,11 @@ export default {
           'progressPercent'] += ((1/this.newProgressBarInfo['totalQuestions'])*100)
       }
 
-      setTimeout(() => {
+      this.$nextTick(() => {
         this.$refs["progressBarRef"].progressTo(
           this.newProgressBarInfo['progressPercent']
         );
-      }, 200)
+      })
```
deepansh96, line 370: `this.$nextTick` removes the need for a `setTimeout`

src/components/PlioQuestion.vue
dalmia: Screenshot 2021-02-28 at 12 01 01 PM
this seems to indicate that this function is called before the repaint?

src/components/PlioQuestion.vue
dalmia, line 273: spacing befoe and after comma

src/components/PlioQuestion.vue
dalmia: not sure if @pritamps knows a better way of doing it?

src/components/PlioQuestion.vue
```diff
@@ -218,13 +227,88 @@ export default {
...
+        optionParents.forEach(option => {
+          var optionTexts = option.querySelectorAll('p')
+          if (optionTexts == null || optionTexts.length == 0) return
+          
+          optionTexts.forEach(optionText => {
+            optionText.style.lineHeight = '100%'
+            optionText.style.marginBlock = "5px"
+          })
+        });
+      })
```
dalmia, line 311: looks very similar to the function for questions - wonder if we can remove this redundancy in some way?

src/components/PlioQuestion.vue
```diff
@@ -218,13 +227,88 @@ export default {
...
+        // paste it right outside, just below it.
+        p_tags.forEach(p_tag => {
+          var img_tag = p_tag.querySelector('img')
+          if (img_tag != null) {
+            p_tag.removeChild(img_tag)
+            p_tag.parentNode.insertBefore(img_tag, p_tag.nextSibling)
+            img_tag.setAttribute("id", "question_image")
+            // scale the image such that width is 1/3rd of screen width
+            // maintaining the aspect ratio
+            var heightToWidth = (img_tag.offsetHeight/img_tag.offsetWidth)
```
dalmia, line 261: better?
```suggestion
            var aspectRatio = (img_tag.offsetHeight/img_tag.offsetWidth)
```

dalmia (changes requested):
Looks great. Minor changes suggested 🚀

src/App.vue
pritamps: Does this work after deploying? Maybe we should put the font on s3, make it public and include the URL here. Or use a CDN or something

src/components/PlioQuestion.vue
pritamps: OPTIONAL: Just a style suggestion. Can we rename `index` to `optionNumber`?

src/components/PlioQuestion.vue
```diff
@@ -47,9 +48,9 @@
               v-if="!isAnAnsweredQuestion && !isTutorialComplete && !tutorialProgress['options']">
             </mcqOptionsPointer>
           </div>
-          </div>
+        </div>
 
-        <!-- revise button -->
+          <!-- revise button -->
```
pritamps, line 53: Why this extra indent?

src/components/PlioQuestion.vue
pritamps: I think the argument (no argument given here) is the function that will be called before the repaint, and the method itself will be called after

pritamps: Actually, reading more, I'm not 100% sure this works. It waits for the next screen repaint, which could actually be before eerything is loaded, for example in the middle of an animation at some frame

src/App.vue
pritamps: Does this work after deploying also, since it's a local file? Might be better to store it on S3 or somewhere

src/components/PlioQuestion.vue
pritamps: OPTIONAL style suggestion: Maybe rename `index` to `optionNumber`?

src/components/PlioQuestion.vue
pritamps: `requestAnimationFrame` can be triggered at the next repaint, not the end of the DOM loading. Unless I'm understanding something wrong!

src/components/PlioQuestion.vue
pritamps: Cool!

src/components/PlioQuestion.vue
pritamps: Why is this needed?

src/components/PlioQuestion.vue
pritamps: Aspect Ratio is width/height (not height/width). but otherwise yes, use aspect ratio!

src/components/PlioQuestion.vue
pritamps: I'm not sure what @deepansh96 is saying is true. if the styles are defined and CSS is ready, they should be loaded whenever the paragraphs are loaded. For example, you're not setting fonts etc here, and they are being read from CSS only.

what I would suggest here is to add a class to each of the `p_tag`s, and define that class's style in the CSS

src/components/PlioQuestion.vue
pritamps: Add class vis JS and set style in the CSS. Separate JS and CSS as far as possible, on principle

src/components/PlioQuestion.vue
```diff
@@ -503,12 +624,13 @@ input {
   &__body {
     padding: 4px 4px 4px;
     padding-left: 16px;
-    margin-bottom: 0;
+    // margin-bottom: 0;
```
pritamps, line 627: Remove this line

pritamps (changes requested):
This is really cool work!!

pritamps:
Sorry for some double comments. Browser window closed by mistake, and I didn't know if the commnets were published or not!

src/App.vue
deepansh96: It works after deploying as the font is packaged along with the frontend files and hosted on S3. The only benefit of keeping the font on S3 at some location would be that the size of the static web files will be much smaller

pritamps: Much? Font is that big kya? 

Might make sense to keep it in s3 and load it only when needed, then?

pritamps: Today is Thursday lol. Deploy it off if the size of the font isn't crazy

src/components/PlioQuestion.vue
deepansh96: Renamed the variable to aspectRatio

pritamps: It's not aspect ratio. Aspect ratio is `width/height`, not `height/width` ... so if you're going with aspect ratio, change teh calculation and use the new value

deepansh96: Oh I thought you said that I should just change the variable name :P

src/components/PlioQuestion.vue
deepansh96: Should I change it back to a setTimeout?
I tested it extensively, it's working in mobile and laptop

pritamps: I don't know. Maybe I'm wrong :D if you are sure what `requestAnimationFrame` does, you can merge! But also be sure.

src/components/PlioQuestion.vue
deepansh96: The `` tag is a child of the `<p>` tag. I took it out and placed it next to the `<p>` tag so I'm able to manipulate it easily.
The parent of `<p>` tag has `flex` display property. To align the image in the center, it needs to be taken out. 
```
flex-direction: column;
align-items: center;
```
The above piece of code helped center the image independent of the `<p>` tag

src/components/PlioQuestion.vue
deepansh96: I tried this. Added classes to each of the `<p>` tags and defined those CSS values in that class.
It didn't work somehow.
While inspecting the rendered HTML code, I was able to see that the class was added to the rendered `<p>` tags but the CSS styles were not applied. The CSS styles that were applied to those `<p>` tags were the default chrome ones.

src/components/PlioQuestion.vue
deepansh96: Done. Committing

src/components/PlioQuestion.vue
pritamps: Hmm... It wasn't possible to insert after the previous `p` tag? 
Anyway, this not a big deal. Can you add a comment in the code?

src/components/PlioQuestion.vue
```diff
@@ -258,9 +258,12 @@ export default {
...
             // scale the image such that width is 1/3rd of screen width
             // maintaining the aspect ratio
-            var aspectRatio = (img_tag.offsetHeight/img_tag.offsetWidth)
+            var aspectRatio = 0
+            if (img_tag.offsetHeight){
+              aspectRatio = (img_tag.offsetWidth / img_tag.offsetHeight)
+            }
             img_tag.style.width = String(window.screen.availWidth/3) + 'px'
-            img_tag.style.height = String(img_tag.offsetWidth * aspectRatio) + 'px'
+            img_tag.style.height = String(img_tag.offsetWidth / aspectRatio) + 'px'
```
pritamps, line 266: This will give you a division by zero error when the aspectRatio is 0.

Also this seems a little unnecessary. The aspect ratio calculation can be avoided and you can simply do:

```js
if (img_tag.offsetHeight) {
    img_tag.style.height = img_tag.offsetHeight
} else {
    img_tag.style.height = 0
}

Becuase you're dividing the `offsetWidth` by the `aspectRatio` and the result of that division is simply

**Review by pritamps — APPROVED** [review]
Quick fix and then you can merge!

File: src/components/PlioQuestion.vue (hunk 17 above)
Comment (deepansh96) [reply src/components/PlioQuestion.vue:266]: Uhhhhh how did i miss this

Comment (pritamps) [reply src/components/PlioQuestion.vue:266]: Yo yo 

I think your calculation is wrong..

It should be 
```
img_tag.style.height = String(img_tag.style.width / aspectRatio) + 'px' # note `width` instead of offsetWidth
```

Otherwise the image will not maintain its AR

deepansh96: No actually I want to scale the image such that it's width is 1/3rd of the screen width.
Otherwise it looks very small inherently

pritamps: Haan I understand. 

The width is 1/3rd, and the height should be set based on two things: the new width (`img_tag.style.width`) and the aspect ratio (`aspectRatio`).

Otherwise you're setting the height to the same as the old height, which is not what you want since image should scale proportionately

deepansh96: isme bass `img_tag.style.width` should be `img_tag.offsetWidth`
offsetWidth gives an integer
so easy to divide directly

dalmia (approved):

PR #75, opened 2021-03-10
WIP: adding language localization - working for one page

### LINKED ISSUE #74: Language localization

dalmia:
Fixes #74 

## Summary

We want to support switching between multiple languages. This PR will make it possible.
- [x] Translate all platform strings
- [x] place the locale switcher in a decent place
- [x] wherever remote config is being updated, update the local config
- [x] if not logged in or logged in, fetch user config and set local config
- [x] Set/update preference in user config (local and remote)
- [x] Fetch default locale from vuex store

## Test Plan
Will start testing once I am done with the changes.

## Resources
Using the following resources as a guide:
- [Guide to localization in Vue](https://phrase.com/blog/posts/ultimate-guide-to-vue-localization-with-vue-i18n/#Creating_our_Demo) - this is for the Vue2 version of the plugin. So, I am tweaking a few things from here but overall a good guide.
- Vue-i18n [documentation](https://vue-i18n.intlify.dev/guide/)


pritamps (changes requested):
Cool!!

Just remove that sample file (HelloI-8n.vue), and is good to go!

pritamps:
Maybe also:

1. have different files for each language, and then combine into one?
2. Add a TODO (not needed now) for updating the files remotely (so the repo has defaults, but the strings come from a service), so that strings can be changed on the fly without deployment/developer time

pritamps:
4. This may or may not be important, but only way to know is to answer it :D: what is the size of the localization library once bundled?

src/App.vue
```diff
@@ -1,14 +1,22 @@
...
 
 <script>
+import LocaleSwitcher from "./components/LocaleSwitcher.vue";
 export default {
+  components: {
+    LocaleSwitcher,
+  },
+  created() {
+    console.log(this.$i18n.availableLocales);
+  },
```
deepansh96, line 19: can remove this if it was just used for local testing?
The console logs are not used when deploying on prod.
If a developer needs to see this info in console, then we can let it be
otherwise remove it

pritamps: Deepansh -- didn't you enable some plugin that removes all console logs before shipping? Or am I remembering wrong?

deepansh96: Yes yes it does remove
But only when we compile for deployment
While development, the console logs will remain

I try to remove them after developing and testing because they clog up the chrome console (in local dev)

None of this will matter when shipping

package.json
```diff
@@ -1,58 +1,60 @@
...
+        "vuex": "^4.0.0-beta.4"
     },
-    "rules": {}
-  },
-  "browserslist": [
-    "> 1%",
-    "last 2 versions",
-    "not dead"
-  ]
+    "devDependencies": {
```
pritamps, line 27: These changes are all by some auto formatter?

src/App.vue
```diff
@@ -1,14 +1,19 @@
 <template>
   <div id="nav">
-    <router-link to="/">Home</router-link> |
-    <router-link v-if="!isLoggedIn" to="/login/">Login</router-link>
-    <a href="#" v-if="isLoggedIn" @click="logout"> Logout</a>
+    <router-link to="/">{{ $t("nav.home") }}</router-link> |
```
pritamps, line 3: Nice :)

src/components/HelloI18n.vue
```diff
@@ -0,0 +1,19 @@
+<template>
+  <p>{{ $t('hello') }}</p>
```
pritamps, line 2: Remove file if not used

src/i18n.js
```diff
@@ -0,0 +1,24 @@
...
+/**
+ * Load locale messages
+ * 
+ * The loaded `JSON` locale messages is pre-compiled by `@intlify/vue-i18n-loader`,
+ * which is integrated into `vue-cli-plugin-i18n`.
+ * See: https://github.com/intlify/vue-i18n-loader#rocket-i18n-resource-pre-compilation
+ */
+function loadLocaleMessages() {
+    const messages = {
+        en: translationsEn,
```
pritamps, line 14: Excellent :)

pritamps (changes requested):
Just asking for one file deleted. This is great!

How does the player decide which language to show, finally? Maybe missed that in the PR

dalmia:
> Maybe also:
> 
> 1. have different files for each language, and then combine into one? Makes it easier to read and edit.
> 2. Add a TODO (not needed now) for updating the files remotely (so the repo has defaults, but the strings come from a service), so that strings can be changed on the fly without deployment/developer time
> 3. Add some error checking for missing strings. But maybe this is for future PRs when we actually add the translations

Agreed. There is a fallback locale option in general for missing strings - which I have kept as English. But yes, makes sense to have both the remote changing of strings and error checking for missing strings 👍

dalmia:
> 1. This may or may not be important, but only way to know is to answer it :D: what is the size of the localization library once bundled?

Size is around 500kb: https://www.npmjs.com/package/vue-i18n

package.json
dalmia: Yep - just after npm install.

src/App.vue
dalmia: Removed this!

dalmia:
> Wow!
> 
> This looks EXTREMELY remote friendly. Can we please please take the extra time and build it out as an API call, and have the strings stored on S3. We can then later add an editor to edit strings.
> 
> These can stay here as default, but while setting the strings for current session, it should try the API call for the current locale.
> 
> This way, strings will be independent of deployment, and more importantly from a philosophy POV, tech team won't have be involved for (for example) small string changes.
> 
> We can upload English strings, and then translations can come from remote place
> 
> Hmm, on second thought, let's not do this right now (unless you want to :P), but can you start tracking it in an issue?

Again, agreed that this would be super helpful! :)

dalmia:
> Just asking for one file deleted. This is great!
> 
> How does the player decide which language to show, finally? Maybe missed that in the PR

We get an instance of `$i18n` accessible throughout our app which has a `.locale` attribute. Whenever that value changes, automatically the language gets changed everything. And that value is reactively changed in `components/LocaleSwitcher.vue`. Does that make sense?

dalmia:
Sharing one issue that I encountered:

In case anyone has any answers, I am encountering a weird bug with Vuex. We are using Vuex with LocalStorage. Once a user logs in, I am fetching the user config and storing it locally on the local storage through Vuex. However, right after Vuex returns as done, if I print the value of localStorage.config -> comes out fine. But if get the same value through Vuex (Vuex also fetches the value through localStorage itself), it shows as null.

However, if I refresh the page once, I get the same value through both the routes
What could be the issue? Is Vuex caching the value somewhere and only after refresh fetching the updated value from the localStorage?

FYI, this issue exists for `localStorage.phone` too. Have made everything work by directly using `localStorage` everywhere but I would have preferred fetching values from Vuex.

dalmia:
Please check my new changes as well - added user configs being fetched, stored locally and locale preference being saved and updated as well - had to do some tweaking to handle those who will login for the first time and those who would be already logged in. I have tested it quite a lot locally but still would love to have a second eye.

pritamps:
> Sharing one issue that I encountered:
> 
> In case anyone has any answers, I am encountering a weird bug with Vuex. We are using Vuex with LocalStorage. Once a user logs in, I am fetching the user config and storing it locally on the local storage through Vuex. However, right after Vuex returns as done, if I print the value of localStorage.config -> comes out fine. But if get the same value through Vuex (Vuex also fetches the value through localStorage itself), it shows as null.
> 
> However, if I refresh the page once, I get the same value through both the routes
> What could be the issue? Is Vuex caching the value somewhere and only after refresh fetching the updated value from the localStorage?
> 
> FYI, this issue exists for `localStorage.phone` too. Have made everything work by directly using `localStorage` everywhere but I would have preferred fetching values from Vuex.

probably cache yes. Looks like you need to update a mutation or something: https://stackoverflow.com/questions/55356977/vuex-state-not-updating-from-local-storage#:~:text=directly%20(for%20example%20clicking%20a,the%20key%20doesn't%20exist.

src/App.vue
```diff
@@ -1,51 +1,121 @@
...
+          this.$store.dispatch("saveConfig", {
+            config: JSON.stringify(response.data), // save user config locally
+          });
+        });
+    },
+    setLocale() {
+      var redirectId = setInterval(() => {
+        // for some reason, the localStorage.config has the correct value
+        // but when fetching the same through Vuex, it gives null
+        // upon refreshing, the fetch from Vues starts working fine
```
pritamps, line 53: Because of this: https://stackoverflow.com/questions/55356977/vuex-state-not-updating-from-local-storage#:~:text=directly%20(for%20example%20clicking%20a,the%20key%20doesn't%20exist.

src/components/LocaleSwitcher.vue
```diff
@@ -0,0 +1,59 @@
...
+      <option value="hi">हिंदी</option>
+    </select>
+  </div>
+</template>
+
+<script>
+export default {
+  name: "LocaleSwitcher",
+  methods: {
+    setLocale() {
```
pritamps, line 14: It sees like we're assuming locale will be switched ONLY on `userConfig`. Just to be a little future-proof, I would suggest having two methods:

```
setLocale() {
// Simply set locale in the app
} 

setLocaleBasedOnUserConfig() {
// Fetch user config
setLocale()
}
```

Make sure all locale changing related logic is in this file only

pritamps: This method is setting the user config, or is it setting the current locale that app will show (which is what the name `LocaleSwitcher` seems to indicate. Little confused!

src/views/PhoneSignIn.vue
```diff
@@ -85,6 +88,35 @@ export default {
         });
     },
 
+    saveLocalUserConfig() {
```
pritamps, line 91: So this component is now doing two things:

1. Signing in with the phone
2. Saving user config, setting Locale etc

I would suggest separating the logic (`PhoneSignIn.vue`, `SetUserProperties.vue`), so that when we implement email sign in/other auth protocols, you just need to implement the sign in logic.

I see some code for experiments, etc, which should also be out of this file.

src/views/Player.vue
```diff
@@ -375,28 +374,32 @@ export default {
...
       if (this.isTutorialUploadRequired) {
-        this.userConfigs["tutorial"]["isComplete"] = this.isTutorialComplete;
-        this.userConfigs["tutorial"]["progress"] = this.tutorialProgress;
-        const tutorial_progress = {
+        this.userConfig["tutorial"]["isComplete"] = this.isTutorialComplete;
+        this.userConfig["tutorial"]["progress"] = this.tutorialProgress;
+        const userConfig = {
           "user-id": this.userId,
-          configs: this.userConfigs,
+          configs: this.userConfig,
```
pritamps, line 381: Continuing renaming, and name `configs` -> `config` also?

pritamps (changes requested):

dalmia:
> > > Just asking for one file deleted. This is great!
> > > How does the player decide which language to show, finally? Maybe missed that in the PR
> > 
> > 
> > We get an instance of `$i18n` accessible throughout our app which has a `.locale` attribute. Whenever that value changes, automatically the language gets changed everything. And that value is reactively changed in `components/LocaleSwitcher.vue`. Does that make sense?
> 
> No I mean -- is there a UI button deciding the locale? Or is it changed as a plio setting?

Yep - you can look at `components/LocaleSwitcher.vue`. It makes a UI button as shown in the image.
Screenshot 2021-03-12 at 10 32 13 AM

src/views/Player.vue
dalmia: The only issue here is that we'll have to go and rename all user config jsons. This is the default structure of the user config - I had initially also proposed keeping `config` here but deepansh felt that because this contains multiple configs like for tutorial progress, experiment allocation and now, locale, it should be plural.

pritamps:
> > No I mean -- is there a UI button deciding the locale? Or is it changed as a plio setting?
> 
> Yep - you can look at `components/LocaleSwitcher.vue`. It makes a UI button as shown in the image.
> Screenshot 2021-03-12 at 10 32 13 AM

Oh.. just thinking aloud here a bit:
1. So we want students to have the ability to change locale? If we want to set language for a student (for example, set Hindi for all students in Haryana), we have to do it from user config? 
2.  Does the button look good on mobile? And does it show on the player screen also in the same place, because there, the video is covering the whole screen no?

src/views/Player.vue
pritamps: Then maybe keep the other one also as plural? Now we are both inconsistent and incorrect :D Better to be one of the two :D

dalmia:
> > > 1. This may or may not be important, but only way to know is to answer it :D: what is the size of the localization library once bundled?
> > 
> > 
> > Size is around 500kb: https://www.npmjs.com/package/vue-i18n
> 
> This is a lot, and will increase our loading time quite a bit :( Can you check in the Vue UI about the overall package dimensions. @deepansh96 had done this when we ditched the phone library I think, so he can help.

I just checked on the Vue UI - this library `vue-i18n` is taking up only 80kb. That is fine, no?
Screenshot 2021-03-12 at 10 37 36 AM

dalmia:
> > > No I mean -- is there a UI button deciding the locale? Or is it changed as a plio setting?
> > 
> > 
> > Yep - you can look at `components/LocaleSwitcher.vue`. It makes a UI button as shown in the image.
> > Screenshot 2021-03-12 at 10 32 13 AM
> 
> Oh.. just thinking aloud here a bit:
> 
> 1. So we want students to have the ability to change locale? If we want to set language for a student (for example, set Hindi for all students in Haryana), we have to do it from user config?
> 2. Does the button look good on mobile? And does it show on the player screen also in the same place, because there, the video is covering the whole screen no?

yep sharing the screenshots - had tested them as well
Screenshot 2021-03-12 at 10 39 27 AM
Screenshot 2021-03-12 at 10 39 09 AM

On the choice of language for the student, I have kept hindi as the default locale to begin with. If the student changes it to English, I think it is their choice and we shouldn't override that. In the future, once we have the creator moved here - the creator can choose the default language for their plios - which is what the learner will see unless they have explicitly set their locale - in which case we will use whatever locale they have set. Does that sound reasonable?

pritamps:
> On the choice of language for the student, I have kept hindi as the default locale to begin with. If the student changes it to English, I think it is their choice and we shouldn't override that. In the future, once we have the creator moved here - the creator can choose the default language for their plios - which is what the learner will see unless they have explicitly set their locale - in which case we will use whatever locale they have set. Does that sound reasonable?

Yes, seems very reasonable to me! 👍🏾 But seems like you're designing without user research? ;-)

pritamps:
> > > On the choice of language for the student, I have kept hindi as the default locale to begin with. If the student changes it to English, I think it is their choice and we shouldn't override that. In the future, once we have the creator moved here - the creator can choose the default language for their plios - which is what the learner will see unless they have explicitly set their locale - in which case we will use whatever locale they have set. Does that sound reasonable?
> > 
> > 
> > Yes, seems very reasonable to me! 👍🏾 But seems like you're designing without user research? ;-)
> 
> Just an initial call (as we have to take an initial call right? :P) - but definitely will be validating it once people start using.

Don't know man. I thought we weren't supposed to make assumptions any more and empathize, ideate, etc before implementing :P 

(just teasing. I think this is the correct choice!)

dalmia:
> > Sharing one issue that I encountered:
> > In case anyone has any answers, I am encountering a weird bug with Vuex. We are using Vuex with LocalStorage. Once a user logs in, I am fetching the user config and storing it locally on the local storage through Vuex. However, right after Vuex returns as done, if I print the value of localStorage.config -> comes out fine. But if get the same value through Vuex (Vuex also fetches the value through localStorage itself), it shows as null.
> > However, if I refresh the page once, I get the same value through both the routes
> > What could be the issue? Is Vuex caching the value somewhere and only after refresh fetching the updated value from the localStorage?
> > FYI, this issue exists for `localStorage.phone` too. Have made everything work by directly using `localStorage` everywhere but I would have preferred fetching values from Vuex.
> 
> probably cache yes. Looks like you need to update a mutation or something: https://stackoverflow.com/questions/55356977/vuex-state-not-updating-from-local-storage#:~:text=directly%20(for%20example%20clicking%20a,the%20key%20doesn't%20exist.

Yayyy - this solves it - we did have mutations in place but we were not updating these values in the mutation. After doing so, it is working. Thank you so much - I feel very happy now.

src/components/LocaleSwitcher.vue
dalmia: Oh, let me clarify this better. So, firstly, we will be fetching and storing the user config locally before any of the language switch - either once they log in (PhoneSignIn.vue) or for logged in users this is fetched in App.vue. Then, changing the `select` box will automatically change the current locale of that app (without any other function). The change in the select box will then invoke this function which will update the user config locally and remotely. So, I don't see a need for 2 different functions unless I have not understood something that you are saying. Does this make sense?

src/views/Player.vue
dalmia: achcha got it. makes sense.

src/views/PhoneSignIn.vue
dalmia: The code for experiments is just to redirect the user to the page where they came from. Nothing else.

dalmia: Looking at the separation of the logic now.

dalmia:
Fixed the config vs configs inconsistency - had to update backend too
Backend PR: https://github.com/avantifellows/plio-backend/pull/53

src/views/PhoneSignIn.vue
dalmia: Done

src/components/LocaleSwitcher.vue
pritamps: Still -- function is doing two things. Maybe rename to `setLocaleAndUpdateUserConfig`. 

Especially since you're updating user config, which sounds like a general operation, not only for locale. Perhaps should be in `UserProperties` component no?

src/components/SetUserProperties.vue
```diff
@@ -0,0 +1,37 @@
...
+            this.$store.getters.getUserId
+        )
+        .then((response) => {
+          this.$store.dispatch("saveConfigs", {
+            configs: JSON.stringify(response.data), // save user config locally
+          });
+        });
+    },
+
+    setLocale() {
```
pritamps, line 25: There is `UserProperties.setLocale` and `LocaleSwitcher.setLocale`. There is some cleanup here -- maybe move that method also here and split by functionality, since both are dealing with the user config?

src/views/Error.vue
```diff
@@ -1,175 +1,177 @@
...
-        <img src="../assets/chrome.svg">
+        class="icon-block"
+      >
+        <img src="../assets/chrome.svg" />
       </a>
       <div class="lead_text">
-        <p>ऊपर click करके Google Chrome download करें</p>
-        <p>या फिर</p>
-        <p>Link कॉपी करके Google Chrome में खोलें</p>
+        <p>&#128070;</p>
```
pritamps, line 86: What is this number here? :D

src/views/PhoneSignIn.vue
```diff
@@ -50,24 +55,31 @@ export default {
...
       })
         .then(() => {
           this.$store.dispatch("login", {
-            phone: this.phone_input,
+            phone: this.phoneInput,
           });
         })
+        .then(() => {
+          // set user config locally
+          this.$refs.userProperties.saveLocalUserConfigs();
```
pritamps, line 75: Nice!

pritamps (changes requested):
Looks great! One change or clarification still needed.. Can clarify and ship also!

src/views/Error.vue
dalmia: Just a finger emoji - pointing up :P

dalmia:
Great feedback on reusing the code components for config and locale @pritamps. Please have a look at my current changes - you might like it! :)

pritamps (approved):
Yeah! I do like a lot 👍🏾 👍🏾 !! Cool Work

I liked it before also, of course :P This is high quality stuff! 

I'm assuming @deepansh96 has looked at it also :P

deepansh96:
All good from my side. Have you already tested on staging or you'll do it after merging?

dalmia:
Tested everything on staging too

> All good from my side. Have you already tested on staging or you'll do it after merging?

Done testing on staging.

deepansh96:
> Tested everything on staging too
> 
> > All good from my side. Have you already tested on staging or you'll do it after merging?
> 
> Done testing on staging.


https://user-images.githubusercontent.com/20169829/110942932-8c5df900-8360-11eb-92ca-d7f0df644d05.mp4

dalmia:
Deployed! 🚀

PR #88, opened 2021-03-13
FIX: don't try updating locale when logged out

### LINKED ISSUE #87: Error when trying to update locale if logged out
**Describe the bug**
Every time someone switches the locale, the code updates  the local and remote user config. However, there is no user config when the user is logged out. Currently, it still tries to update them which gives an error.

dalmia:
Fixes #87 

## Summary
Only update locale if the user is logged in.

## Test Plan
Tested it locally - now, if logged out, no error is shown.

### DIFF
--- a/src/components/UserProperties.vue
+++ b/src/components/UserProperties.vue
@@ -34,12 +34,14 @@ export default {
     },
 
     updateLocale() {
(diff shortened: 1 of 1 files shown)


pritamps (commented):
Maybe hide the button also if not logged in?

pritamps (approved):
Oh never mind - understood :)

PR #93, opened 2021-03-15
FIX: make home and login/logout clickable again

### LINKED ISSUE #92: Home and Logout button are not clickable due to overlay
Try using z-index negative

dalmia:
Fixes #92 

## Summary
Remove `overlay` and use `display=flex` with hack of showing the locale switcher on both sides and making the one on the left invisible. Couldn't figure out any other way of doing this. 

## Test Plan
Tested locally.

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -1,11 +1,14 @@
 <template>
   <div id="header">
+    <div>
+      <LocaleSwitcher id="locale" class="hidden"></LocaleSwitcher>
+    </div>
     <div id="nav">
       <router-link to="/">{{ $t("nav.home") }}</router-link> |
       <router-link v-if="!isLoggedIn" to="/login/">{{ $t("nav.login") }}</router-link>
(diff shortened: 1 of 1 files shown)


src/App.vue
```diff
@@ -1,11 +1,14 @@
 <template>
   <div id="header">
+    <div>
+      <LocaleSwitcher id="locale" class="hidden"></LocaleSwitcher>
```
deepansh96, line 4:
```suggestion
      <LocaleSwitcher id="locale" v-show="false"></LocaleSwitcher>
```

deepansh96 (changes requested):
Just one change suggested, can merge after that

src/App.vue
dalmia: Nice :+1:

dalmia: This removes the element - does not make it invisible.

deepansh96 (approved):

PR #94, opened 2021-03-15
FIX: added change to keep position of switcher in player too

### LINKED ISSUE #92: Home and Logout button are not clickable due to overlay

dalmia:
Fixes #92 

## Summary
Just a follow up to #93 - position of the switcher changed for the player. This fixes that. 

## Test plan
Tested locally and on staging - for home and player

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -1,6 +1,6 @@
 <template>
   <div id="header">
-    <div>
+    <div class="left">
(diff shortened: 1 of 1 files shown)


deepansh96 (approved):

dalmia:
Merged and deployed.

PR #101, opened 2021-03-16
Fixes plio question and answer structure 

### LINKED ISSUE #66: Fix question object data structure
Within a plio JSON, ideally there should be a list of `items`. Currently, it is a list of `questions`. Also, the key for the correct answer is `answers` - which is weird.

dalmia:
Fixes #66 

## Summary
Handles the new plio question structure and changes the structure of the answers. More details here: https://github.com/avantifellows/plio-maker/issues/21

## Test Plan
Tested it locally with some test plios I have created. There are some TODOs after the PR gets merged which can also be found listed in the page linked above.


src/components/PlioQuestion.vue
```diff
@@ -6,23 +6,29 @@
       <div class="modal__dialog">
         <div class="modal__body">
           <!-- The cross button to close the modal -->
-          <div class="question_text_row">
-            <div class="question_text" id="question" v-html="questionText"></div>
+          <div class="question-text-row">
+            <div class="question-text" id="question" v-html="questionText"></div>
             <div class="close-container" id="skip-button" @click="clickSkip">
+              <!-- {{ isAnswerSubmitted }} -->
```
deepansh96, line 12: Is this needed?

src/components/PlioQuestion.vue
```diff
@@ -212,37 +214,35 @@ export default {
...
       // Show highlighted options if coming back to an answered question
       // Wait 200 ms because it takes some time to find the DOM elements
-      if (this.plioQuestion.state == "answered") {
+      if (this.isAnAnsweredQuestion) {
         requestAnimationFrame(() => {
+          this.isAnswerSubmitted = true;
+
           // highlight wrong/right depending on what the user answered in previous session
-          document.getElementById("options_container").classList.add("options-block");
+          // document.getElementById("options-container").classList.add("options-block");
```
deepansh96, line 233: can remove this right?

src/components/PlioQuestion.vue
```diff
@@ -212,37 +214,35 @@ export default {
...
           var selectedOption = document.getElementById(
-            `option_input_${this.userAnswerIndex}`
+            `option_input_${this.plioQuestion.userAnswer}`
           );
           selectedOption.checked = true;
-          this.selectedOption = this.plioQuestion.user_answer;
-          this.checkAnswer();
+          this.selectedOption = this.plioQuestion.item.details.options[
+            this.plioQuestion.userAnswer
```
deepansh96, line 239: Can we mention somwhere that `userAnswer` is actually an index and not the text?

src/components/PlioQuestion.vue
```diff
@@ -315,8 +314,8 @@ export default {
     renderContent() {
       // force rendering of question and option texts
       this.$nextTick(() => {
-        MathJax.Hub.Queue(["Typeset", MathJax.Hub, "question"]);
-        MathJax.Hub.Queue(["Typeset", MathJax.Hub, "option"]);
+        // MathJax.Hub.Queue(["Typeset", MathJax.Hub, "question"]);
+        // MathJax.Hub.Queue(["Typeset", MathJax.Hub, "option"]);
```
deepansh96, line 318: You forgot to uncomment

src/components/PlioQuestion.vue
```diff
@@ -496,6 +474,13 @@ export default {
       "https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.7/MathJax.js?config=TeX-MML-AM_CHTML";
     document.head.appendChild(mathJaxCDN);
   },
+  emits: [
+    "update-journey",
+    "question-skipped",
+    "answer-submitted",
+    "clicked-revise",
+    "question-completed",
+  ],
```
deepansh96, line 483: why is this required?

dalmia: Vue gives a warning.

deepansh96 (approved):

src/views/Player.vue
```diff
@@ -229,7 +229,7 @@ export default {
         )
         .then((res) => {
           console.log(res.data);
-          var questions = res.data.plioDetails.questions.questions;
+          var questions = res.data.plioDetails.items;
```
pritamps, line 232: Shouldn't you check for the type of items here? Questions should be only those that have type `question` no?

src/views/Player.vue
```diff
@@ -229,7 +229,7 @@ export default {
         )
         .then((res) => {
           console.log(res.data);
```
pritamps, line 231: Remove old log statement :D

src/views/Player.vue
```diff
@@ -262,13 +262,13 @@ export default {
             let plioQuestion = {
```
pritamps, line 262: This is a very strange loop -- hadn't noticed this before. You can replace it with a `questions.forEach` loop and it will look cleaner, I think....

Maybe do it off in this PR only

pritamps (changes requested):
Great that this is happening.. Just a few things!

src/views/Player.vue
dalmia: Yep, will do! 👍

src/views/Player.vue
dalmia: I haven't done that currently intentionally because we don't know how other items could look like and correspondingly, how should answers look then. Since I was a bit unsure of how any new items could look like, I chose to keep this as is and make that specific change when we do end up adding another item type. What do you think?

pritamps: Questions should still be only those that have type question, IMO. Rest of items will probably end up having their own component, I think!

dalmia:
@pritamps incorporated your feedback and left one comment. Please have a look!

src/views/Player.vue
dalmia: Yep. And I think this variable itself will become `items` in the future and we'll either use PlioQuestion or another component as necessary!

pritamps: I know this is merged, but you should rename the variable also to `items` and then just set `questions` to items. The important thing here is that the frontend follows the structure of the JSON as closely as possible. This is what will prevent bugs in the future.

dalmia: Makes sense. Will do 👍

Issue #102, opened 2021-03-16
Small optimisations, encapsulation, loose coupling and cleanup


deepansh96:
**Describe the task**
The current frontend code was not utilising the full power Vue provides. This PR #103 solves some of those issues. 
Some of the tasks are outlined below with their explanation and reference links (if any).

- [x] Encapsulate axios API calls into separate service files. Attaching a quote from the vuemastery course,
> If we’re importing Axios into each component that needs it, we’re unnecessarily creating a new instance of Axios each time we do that. With API code woven throughout our application, this gets messy and makes our app harder to debug. A cleaner and more scalable solution is to modularize our API code into a service layer.
- [x] Router links had a tight coupling with the path of the component that it is linking to. This tight coupling is not scalable. If in future, any of our routing path changes, the router link will need to be edited.
- [x] Also, another instance of tight coupling is how params and queries are passed through the router to different components. Instead of accessing them like `this.$route.params.id`, the `id` param should be passed as a prop to the component.
- [x] The loading spinner is added to individual components and is tightly coupled with that component itself. But we also need a loading spinner that works globally across the app.
- [x] While importing files into components and other files, the file paths look like `../../bla_bla.vue`. Using `@/bla_bla.vue` is a better solution. `@` here means `src/`
- [x] Throughout the code, the Vuex store states are being accessed through store getters, which is not required. Store getters should be used if we need a derived state/computed state.
- [x] Utilizing helper functions like `mapState`, `mapActions`, `mapGetters` throughout the code. This significantly reduces the code. Instead of writing `this.$store.dispatch('action-name')`, using a `mapActions` would reduce that line to `this.action-name()`
- [x] In a vuex store module `"plioItems"`, namespacing has to be added.

pritamps:
Excellent!!!

PR #103, opened 2021-03-16
Small optimisations, encapsulation, loose coupling and cleanup 

deepansh96:
Fixes for #102

## Test Plan
Have tested changes locally
Will test them on staging and prod soon


src/App.vue
```diff
@@ -1,26 +1,30 @@
 <template>
   <div id="header">
     <div id="nav">
-      <router-link to="/">{{ $t("nav.home") }}</router-link> |
-      <router-link v-if="!isLoggedIn" to="/login/">{{ $t("nav.login") }}</router-link>
-      <a href="#" v-if="isLoggedIn" @click="logout">{{ $t("nav.logout") }}</a>
+      <router-link :to="{ name: 'Home' }">{{ $t("nav.home") }}</router-link> |
+      <router-link v-if="!isLoggedIn" :to="{ name: 'Phone Sign In' }">{{ $t("nav.login") }}</router-link>
+      <a href="#" v-if="isLoggedIn" @click="logoutUser">{{ $t("nav.logout") }}</a>
```
deepansh96, line 6: https://router.vuejs.org/guide/essentials/named-routes.html

src/App.vue
```diff
@@ -31,16 +35,17 @@ export default {
     this.$refs.userProperties.setLocaleFromUserConfig();
   },
   methods: {
-    logout() {
-      this.$store.dispatch("logout").then(this.$router.push("/login/"));
-    },
+    ...mapActions(['logout']),
```
deepansh96, line 38: https://vuex.vuejs.org/guide/state.html#object-spread-operator

src/components/PlioList.vue
```diff
@@ -3,25 +3,23 @@
...
   name: "PlioList",
   components: {
     PlioThumbnail,
   },
-  computed: mapGetters(["allPlios"]),
-  methods: {
-    ...mapActions(["fetchPlios"]),
-  },
+  computed: mapState('plioItems', ['allPlios']),
+  methods: mapActions('plioItems', ['fetchPlios']),
```
deepansh96, line 22: https://vuex.vuejs.org/guide/modules.html#binding-helpers-with-namespace

src/router/index.js
```diff
@@ -16,20 +16,25 @@ const routes = [{
...
+            id: route.params.id 
+        })
     },
     {
         // type: the type of component invoking this path (optional)
         // id: the unique ID for the component invoking this path (optional)
         path: "/login/:id?/:type?",
         name: "Phone Sign In",
         component: PhoneSignIn,
+        props: true
```
deepansh96, line 30: https://router.vuejs.org/guide/essentials/passing-props.html#passing-props-to-route-components

src/views/Player.vue
```diff
@@ -98,11 +101,14 @@ export default {
       default: "",
       type: String,
     },
+    id: {
+      default: "",
+      type:String
```
dalmia, line 106:
```suggestion
      type: String
```

src/views/PhoneSignIn.vue
```diff
@@ -36,63 +40,51 @@ export default {
...
+            if (this.id) {
+              if (this.type && this.type == "experiment") {
+                // redirect to experiment
+                this.$router.replace({ name: 'ABTesting', params: { id: this.id }});
               } else {
-                this.$router.push({ path: "/" });
+                // redirect to plio
+                this.$router.replace({ name: 'Player', params:{ id: this.id }});
               }
-              clearInterval(redirectId);
```
dalmia, line 92: Test login a few more times ki it works in one shot

dalmia (changes requested):
This is awesome! 🔥 
Left minor comments/suggestions!

dalmia (approved):
LGTM 🚀

src/components/UserProperties.vue
```diff
@@ -3,39 +3,36 @@
...
-            process.env.VUE_APP_BACKEND_USER_CONFIG +
-            "?user-id=" +
-            this.$store.getters.getUserId
-        )
-        .then((response) => {
-          this.$store.dispatch("saveConfigs", {
-            configs: JSON.stringify(response.data), // save user config locally
-          });
-        });
+      UserService.getUserConfig(this.userId)
```
pritamps, line 17: Wah, fancey!

src/components/UserProperties.vue
```diff
@@ -46,35 +43,25 @@ export default {
 
     updateUserConfigs(userConfigs) {
       const jsonUserConfig = JSON.stringify({
```
pritamps, line 45: Why not move this logic also to to the `UserService` and just let `UserService.updateUserConfig` handle all the stuff? Same for other methods

src/components/UserProperties.vue
```diff
@@ -3,39 +3,36 @@
...
+import { mapState, mapActions } from 'vuex'
 
 export default {
+  computed: mapState(['isLoggedIn', 'userId', 'configs']),
   methods: {
+    // object spread operator
+    // https://vuex.vuejs.org/guide/state.html#object-spread-operator
+    ...mapActions(['saveConfigs']),
+
     saveLocalUserConfigs() {
```
pritamps, line 16: Should be in `UserService`?

src/store/index.js
```diff
@@ -4,70 +4,70 @@ import plioItems from './modules/plioItems'
...
             state.userId = null;
         },
-        [CONFIG](state) {
-            state.configFetchedPending = true;
-        },
         [CONFIG_SUCCESS](state) {
-            state.configFetchedPending = false;
             state.configs = localStorage.getItem('configs');
+        },
+        [START_LOADING](state) {
```
pritamps, line 30: Interesting!

pritamps (changes requested):
Looks great! Resident Vue expert 🗡️

src/components/UserProperties.vue
deepansh96: This can't be moved to UserService.js because here, a Vuex action is being dispatched which cannot be done from a JS file. It can only be done from a component

src/components/UserProperties.vue
deepansh96: Same comment as below

src/components/UserProperties.vue
pritamps: Hmm... what is the action that is being dispatched?

1. So then having the `UserService` does't really make sense. If it's not containing all the logic, and the `User` logic is spread across multiple files, no point
OR
2. Design it so that the vuex action is separated from the logic
OR
3. Make `UserService` a component? 

Note that I don't really know which one of the above is possible/feasible/etc. My main point here is that if you're creating a `UserService` class/file, then that should be responsible for `User` stuff, no?

pritamps: https://stackoverflow.com/questions/46716709/how-to-dispatch-action-from-a-javascript-file-instead-of-a-vue-component

Maybe this will work? 

OR this: https://forum.vuejs.org/t/how-to-access-vuex-store-from-a-js-module/10124 

?

deepansh96: I will try these out
The point of having the `UserService` file was to handle and encapsulate the api calls
I agree that this logic should not be dangling around.
Ideally the user config should be fetched by store and saved there itself
and any component can access the stored config by accessing Vuex


I will try these out, until then,I'll merge this PR as Vaibhav has to put out a release

PR #130, opened 2021-03-29
Fixed syncing issues with create new plio

dalmia:


src/App.vue
```diff
@@ -137,8 +137,9 @@ export default {
     },
     onClose(event) {
       // invoked when trying to close the browser or changing pages
-      event.preventDefault();
-      event.returnValue = "";
+      console.log(event);
+      // event.preventDefault();
+      // event.returnValue = "";
```
deepansh96, line 142: we're not keeping that browser popup when navigating away?

src/components/UI/Alert/DialogBox.vue
```diff
@@ -93,25 +96,39 @@ export default {
     },
   },
   computed: {
+    isButtonVisible() {
+      // is either of the confirm of cancel buttons visible
```
deepansh96, line 100: unclear comment

src/components/UI/Alert/DialogBox.vue
```diff
@@ -93,25 +96,39 @@ export default {
...
+      return (
+        this.confirmButtonClass.indexOf("hidden") == -1 &&
+        this.cancelButtonClass.indexOf("hidden") == -1
```
deepansh96, line 103: should also add 1-2 lines explaining this logic?

src/pages/Editor.vue
```diff
@@ -1,104 +1,125 @@
...
+            :iconConfig="backButtonIconConfig"
+            :buttonClass="backButtonClass"
+            @click="returnToHome"
+          ></icon-button>
+          <icon-button
+            :titleConfig="publishButtonTitleConfig"
+            :class="publishButtonClass"
+            v-tooltip.right="publishButtonTooltip"
+            @click="publishButtonClicked"
+          ></icon-button>
```
deepansh96, line 58: also mention in one line what these buttons are
like comments "video link", "item editor", "video preview"

src/pages/Test.vue
```diff
@@ -1,22 +1,19 @@
...
+    SimpleBadge,
   },
-  methods: {
-    confirmClicked() {
-      console.log("confirm");
-    },
+  data() {
+    return {
+      badgeClass: "text-indigo-500 border-indigo-500",
+    };
```
deepansh96, line 16: not removing this?

deepansh96 (changes requested):

deepansh96:
- [ ] when dialog popup appears - it should be always visible on the screen. One should not be able to scroll past it. This is majorly an issue in mobile mode, where most content will be accessible during scrolling. Suppose a user is scrolled on the lower part of the screen, editing a plio. When they click publish, the screen goes blur and no dialog or popup is visible because its stuck to the top of the screen. If you go to primevue site in mobile mode and do this, there the behaviour is what is expected
- [ ] when in mobile mode, i am not able to interact with the marker in any way. I can interact with the player slider and move it around, but any interaction (clicking or dragging) is not possible with the markers
- [ ] in both mobile/desktop mode, after the dialog box appears and the screen blurs, I should not be able to click elsewhere on the screen. At least i should not be able to click on anything that is blurred. But while the dialog box is up there, I am able to click home button, change the player ka slider and do other interactions. I am even able to change the question text in the blurred background
- [ ] in mobile screen, when the dialog box appears, I'm not able to interact with the dialog buttons (yes, no). I cant click them but I can click the home button, question nav buttons etc. So i'm guessing it's a problem with the dialog buttons specifically
- [ ] Also in one video if you'll see, you can see the video coming on top of the toast message whereas it should have been the opposite. Not sure if that has been fixed with toast position recent commit.
- [ ] Same issue with the toast message. If I scroll down, the toast message is attached to the top and goes out of the screen. If you see the video attached below of primevue, the toast should persist no matter where I scroll - but this is low priority as the only toast we have right now is the copy link. But when we start using more toasts, we need them to be visible everywhere.

https://user-images.githubusercontent.com/20169829/113033765-ff63ce00-91ae-11eb-8ab4-6e412b1ae089.mov



https://user-images.githubusercontent.com/20169829/113032963-1b1aa480-91ae-11eb-8df7-76f77c2f1121.mov


https://user-images.githubusercontent.com/20169829/113032588-a34c7a00-91ad-11eb-9772-81330f6198eb.mov


https://user-images.githubusercontent.com/20169829/113034462-b95b3a00-91af-11eb-99bd-da82399bd7f3.mov

dalmia:
Toast issue is sorted

deepansh96 (approved):

PR #132, opened 2021-03-31
WIP: option add and option delete

dalmia:
- [x] support for deleting option
- [x] resets correct option to first option if correct option was deleted
- [x] does not allow deleting option if question has only 2 options - shows error toast
- [x] shows dialog box in ItemEditor before deleting
- [x] fixed sideIcon to startIcon in TextNumber, Textarea and TextInput components
- [x] Added new endIcon in TextInput
- [x] Added some missing documentation stuff for old components


dalmia:
Created PR to master by mistake

PR #135, opened 2021-04-02
Disable add question button when no video URL

dalmia:


src/components/Editor/ItemEditor.vue
```diff
@@ -187,8 +187,8 @@ export default {
       default: 0,
       type: Number,
     },
-    isPublished: {
-      // whether the plio has been published or not
+    isDisabled: {
+      // whether it is allowed to interact with the item editor
```
deepansh96, line 191: we're not showing the editor when it's disabled. Better to make it clear in the comment
"whether to show the item editor"

deepansh96 (changes requested):

PR #136, opened 2021-04-04
REST API integration into frontend

deepansh96:
## Summary
Will be using this PR for integrating the REST API into the frontend

For a list of open questions, proposed changes to the API from the backend side, more endpoint requests etc, please use the "Questions" section in [this](https://www.notion.so/avantifellows/Backend-REST-API-8fe1256783dc43e09d4de5076bfd38eb) doc to maintain that


.env
```diff
@@ -10,3 +10,6 @@ VUE_APP_BACKEND_UPDATE_USER_CONFIG=/users/update_config
 VUE_APP_BACKEND_LOGIN_USER=/users/login
 VUE_APP_I18N_LOCALE=hi
 VUE_APP_I18N_FALLBACK_LOCALE=en
+
+VUE_APP_BACKEND_NEW=http://0.0.0.0:8001/api/v1/
+VUE_APP_BACKEND_PLIO_NEW=plios/
```
deepansh96, line 15: Temporary variables for now, will replace the old ones when as we go on

src/components/UI/PlioThumbnail.vue
```diff
@@ -1,13 +1,13 @@
 <template>
   <!-- named routes - https://router.vuejs.org/guide/essentials/named-routes.html -->
-  <router-link :to="{ name: 'Player', params: { id: plio.plio_id } }" tag="div" class="plio">
+  <router-link :to="{ name: 'Player', params: { id: plio.uuid } }" tag="div" class="plio">
```
deepansh96, line 3: TODO - routing to the player using the `uuid` instead of the `id`

src/services/API/Plio.js
```diff
@@ -24,7 +24,14 @@ export default {
 
   getAllPlios() {
     // returns the list of all plios
-    return apiClient().get(process.env.VUE_APP_BACKEND_PLIOS_LIST);
+    return new Promise((resolve, reject) => {
+      apiClient()
+        .get(process.env.VUE_APP_BACKEND_PLIO_NEW)
+        .then((response) => {
+          if (response.status == 200) resolve(response.data);
+          else reject(response);
```
deepansh96, line 32: TODO - handle the reject section properly

src/components/UI/PlioThumbnail.vue
dalmia: is this still a TODO?

.env
dalmia: need to overwrite older one

src/store/modules/plioItems.js
```diff
@@ -16,8 +16,8 @@ const actions = {
     // exist here, it exists in the root store file "index.js"
     dispatch("startLoading", null, { root: true });
 
-    const response = await PlioService.getAllPlios();
-    var allPlios = response.data["all_plios"];
+    var allPlios = await PlioService.getAllPlios();
+    // var allPlios = data;
```
dalmia, line 20: can remove this right?

src/services/API/RootClient.js
```diff
@@ -1,7 +1,7 @@
 import axios from "axios";
 
 const client = axios.create({
-  baseURL: process.env.VUE_APP_BACKEND,
+  baseURL: process.env.VUE_APP_BACKEND_NEW,
```
dalmia, line 4: need to update this!

src/services/API/Plio.js
dalmia: why can't we just do the `.get` as before? why do we need to return a promise like this?

dalmia (changes requested):
Some comments / questions

src/components/UI/PlioThumbnail.vue
deepansh96: Yes
Right now it's just passing the uuid as a prop to the player component
The component itself needs to be configured with this prop

dalmia (approved):

PR #143, opened 2021-04-07
integrated REST APIs into Editor

deepansh96:
## Summary
- [x] Added endpoints for videos and questions
- [x] Removed old unused endpoints - `plios_list` `get_plio` `create_plio`
- [x] Fixed bug that didn't show the dialog popup after publish action was approved by the user
- [x] Integrated REST APIs for creating/updating `item`, `question`, `video`, `plio`

## Test Plan
Tested locally


.env
```diff
@@ -3,10 +3,10 @@ VUE_APP_FRONTEND=localhost:8080
 
 VUE_APP_BACKEND_PLIOS=plios/
 VUE_APP_BACKEND_ITEMS=items/
+VUE_APP_BACKEND_VIDEOS=videos/
```
rathorevaibhav, line 6: I think we can remove these api endpoints from here and instead keep a config file (maybe config.js) as these endpoints are unlikely to change.

`.env` file should only contain variables that are supposed to differ on various environments.

src/services/API/Video.js
```diff
@@ -0,0 +1,20 @@
+import { apiClient } from "@/services/API/RootClient.js";
+
+export default {
+  async createVideo(videoDetails) {
+    // API to create a video object in the DB
+    var response = await apiClient().post(
+      process.env.VUE_APP_BACKEND_VIDEOS,
+      videoDetails
+    );
+    return response.data.id;
```
rathorevaibhav, line 10: any reason for returning only the id and not the whole object?

I think these service files should return the response we're getting and let the caller function decide what data it needs. That will help making these service functions more generic.

src/services/API/Video.js
```diff
@@ -0,0 +1,20 @@
...
+  },
+
+  updateVideo(videoDBId, videoDetails) {
+    // API to update a video object in the DB
+    return apiClient().patch(
+      process.env.VUE_APP_BACKEND_VIDEOS + videoDBId + "/",
```
rathorevaibhav, line 16: is this still needed or have we configured the trailing slashes to false?

src/services/API/Item.js
```diff
@@ -0,0 +1,24 @@
...
+      .then((response) => {
+        response.data.time = response.data.time / 1000;
+        return response.data;
+      });
+  },
+
+  updateItem(itemDetails) {
+    // API to update an item in the DB
+    itemDetails.time = itemDetails.time * 1000;
+    return apiClient().put(
```
rathorevaibhav, line 19: so there are two types of updates DRF provides:
1. Partial update (uses PATCH)
2. Update (uses PUT)

If we're just updating a few fields (and not all), maybe we can consider the first one.

For example, like below, for session-answer, I'm just updating the answer column and nothing else using "PATCH".
image

.env
deepansh96: Agreed, I remember you suggested this before as well
Will do it

src/services/API/Item.js
deepansh96: Makes sense. My thinking was that in this function, both the things are possible.
Updating the whole schema or just one field
Should I go ahead and create a separate function, something like `partialUpdateItem`?
Or pass an argument here like `isPartial` and the request method switches according to that?

src/services/API/Video.js
deepansh96: Agreed. This was specific to my use case. Will update it!

src/services/API/Video.js
deepansh96: Tried the flag APPEND_SLASH=False in settings.py but was not able to get the desired behaviour. Endpoints were giving Page not found error, so didn't go ahead with that

.env
```diff
@@ -3,10 +3,10 @@ VUE_APP_FRONTEND=localhost:8080
...
+VUE_APP_BACKEND_QUESTIONS=questions/
+
 
-VUE_APP_BACKEND_PLIOS_LIST=/plios_list
-VUE_APP_BACKEND_PLIO_DETAILS=/get_plio
```
dalmia, line 8: this should not be removed in this PR - will be removed in my PR because this is used for Player too.

src/services/API/Video.js
```diff
@@ -0,0 +1,20 @@
...
+    var response = await apiClient().post(
+      process.env.VUE_APP_BACKEND_VIDEOS,
+      videoDetails
+    );
+    return response.data.id;
+  },
+
+  updateVideo(videoDBId, videoDetails) {
+    // API to update a video object in the DB
+    return apiClient().patch(
```
dalmia, line 15: why is this not `.put` instead of `.patch`?
This answer says that `PATCH` is for partial updates: https://stackoverflow.com/questions/31089221/what-is-the-difference-between-put-post-and-patch

Based on the answer, PUT seems more appropriate!

src/services/API/Item.js
dalmia: I think what we need to do is just use PUT and not PATCH in all update functions for now - this is an optimization which we can return to later - can be made an issue - seems very low priority to me.

src/pages/Editor.vue
```diff
@@ -156,6 +156,8 @@ import SliderWithMarkers from "@/components/UI/Slider/SliderWithMarkers.vue";
 import VideoPlayer from "@/components/UI/Player/VideoPlayer.vue";
 import ItemEditor from "@/components/Editor/ItemEditor.vue";
 import PlioService from "@/services/API/Plio.js";
+import ItemService from "@/services/API/Item.js";
+import QuestionService from "@/services/API/Question.js";
```
dalmia, line 160: I have changed this (what I am going to suggest) in my PR but I would recommend using ItemAPIService and QuestionAPIService and in general,  {Resource}{Utility}Service as the name everywhere as there could be other folders which have JS files with the same name - segregated by Resource.

src/pages/Editor.vue
```diff
@@ -811,24 +829,42 @@ export default {
...
-      this.markItemSelected(this.currentItemIndex);
+      // create item, then create the question, then update local states
+      ItemService.createItem({
+        plio: this.plioDBId,
+        type: this.getItemTypeForNewItem(),
+        time: newTimestamp,
+        meta: this.getMetadataForNewItem(),
+      })
+        .then((createdItem) => {
+          newItem = createdItem;
```
dalmia, line 849: what is the use of this line?

src/pages/Editor.vue
```diff
@@ -811,24 +829,42 @@ export default {
...
+              type: "mcq",
+              options: ["", ""],
+              correct_answer: "0",
+            });
+          }
+        })
+        .then((createdQuestion) => {
+          newItem.details = createdQuestion;
+          // push it into items, update the itemTimestamps and currentItemIndex
+          this.items.push(newItem);
```
dalmia, line 863: this does not care about the ordering of the items based on timestamp

src/pages/Editor.vue
```diff
@@ -811,24 +829,42 @@ export default {
...
+        .then((createdItem) => {
+          newItem = createdItem;
+
+          if (createdItem.type == "question") {
+            return QuestionService.createQuestion({
+              item: createdItem.id,
+              type: "mcq",
+              options: ["", ""],
+              correct_answer: "0",
+            });
```
dalmia, line 857: this should come from the functionality of add question

src/services/API/Plio.js
```diff
@@ -1,4 +1,7 @@
 import { apiClient } from "@/services/API/RootClient.js";
+import VideoService from "@/services/API/Video.js";
+import ItemService from "@/services/API/Item.js";
+import QuestionService from "@/services/API/Question.js";
```
dalmia, line 4: same comment on naming

src/services/API/Plio.js
```diff
@@ -48,13 +52,57 @@ export default {
...
-      plio_id: plioId,
-      bucket: "plio-test-data",
-    };
-    return apiClient().post(process.env.VUE_APP_BACKEND_CREATE_PLIO, plioData);
+
+  async updatePlio(value, plioId) {
+    // handle items and questions being updated
+    value.items.forEach((item) => {
+      // pass a deepclone and not the reference
+      var cloneDeep = require("lodash.clonedeep");
```
dalmia, line 60: can be imported globally once outside this loop

src/services/API/Plio.js
```diff
@@ -48,13 +52,57 @@ export default {
...
+    // handle (plio_title, status, created_by) being updated
+    plioObject.name = value.plio_title;
+    plioObject.status = value.status;
+    plioObject.created_by = 1; // TODO - HARDCODED FOR NOW
+
+    return new Promise((resolve) => {
+      // handle video url/duration being updated
+      var isVideoLinked = value.videoDBId != null;
+      var isVideoURLEmpty = value.video_url == "";
+      if (!isVideoURLEmpty) {
```
dalmia, line 78: this check should be made in Editor.vue in checkAndSavePlio

src/services/API/Plio.js
```diff
@@ -48,13 +52,57 @@ export default {
...
+        // create video and save the video db id
+        if (!isVideoLinked) {
+          VideoService.createVideo(videoDetails).then((createdVideoId) => {
+            plioObject.video = createdVideoId;
+            resolve(
+              apiClient().put(
+                process.env.VUE_APP_BACKEND_PLIOS + plioId + "/",
+                plioObject
+              )
+            );
```
dalmia, line 92: this is the same call as below - should not be redundant.

src/services/API/Plio.js
```diff
@@ -48,13 +52,57 @@ export default {
...
+      QuestionService.updateQuestion(itemClone.details);
+    });
+
+    // prepare the payload for plio
+    var plioObject = {};
+
+    // handle (plio_title, status, created_by) being updated
+    plioObject.name = value.plio_title;
+    plioObject.status = value.status;
+    plioObject.created_by = 1; // TODO - HARDCODED FOR NOW
```
dalmia, line 72: Screenshot 2021-04-07 at 9 04 10 PM
I am converting in Editor.vue to the S3 format and you are converting from S3 format to DB format. We can directly go from JS format to DB format - and this should be done in the same place where the DB to S3 format conversion is happening in Editor.vue

dalmia, line 72: let's hard code created_by in vue files only so that we don't have to touch JS files later.

dalmia (changes requested):
Requested some changes.

src/pages/Editor.vue
deepansh96: First we create an item, save the details into `newItem`, then we create a question and save the details into `newItem.details`.
The object that is being passed in the `then` handlers are item details and question details that have been created by the API calls. 
Combining both of those into one object, that is being done by using `newItem`
I can add comments to explain this more clearly. If a better name for the object comes to mind, do suggest

src/pages/Editor.vue
deepansh96: Sorting is being taken care of here. Pushing into `this.items` does not sort inherently, but as it goes to the next line,
```
this.itemTimestamps = this.getItemTimestamps(this.items);
```
the `itemTimestamps` array is updated - still unsorted.

There is a `watch:` method on `itemTimestamps` - 
```
itemTimestamps() {
      this.itemTimestamps.forEach((itemTimestamp, index) => {
        this.items[index]["time"] = itemTimestamp;
      });
      // handle item sorting and marker positioning
      // when time is changed from the time input boxes
      // or when item is added using the add item button
      this.handleTimeUpdateFromEditor();
    },
```
Whenever `itemTimestamps` change, the sorting is handled by the function `handleTimeUpdateFromEditor()`

Although I agree the name of the function can be changed. This function in general sorts the items according to the new timestamps and resets the `currentItemIndex`. Please suggest a better name

dalmia: Got it. My bad. 
We can use `checkAndFixItemOrder`

dalmia:
- update both backend and frontend to work with only seconds - discard milliseconds - don't need to use cloneDeep anymore.

deepansh96:
- [x] Removed the static endpoints from `.env` file and moved them to `src/services/API/Endpoints.js`
- [ ] Returning the whole `video` object instead of just the `video.id` in `Video.js` - Vaibhav's second comment
- [ ] Using `PUT` and not `PATCH` everywhere - will optimise with `PATCH` later - made an issue to track this #145
- [ ] Keeping the unused old API endpoints - will be removed after aman's PR - created an issue for this - #146
- [ ] Changed imported API function names according to Aman's suggestion. Eg - from `import PlioService` to `import PlioAPIService`
- [ ] While filtering and retrieving the items for a particular plio, added an `order_by` on the backend serialiser to send the sorted by time list of items.
- [ ] While creating a new question, earlier the values (two empty options, empty text, correct answer 0) was being hardcoded -- now it's returning from a function
- [ ] `clonedeep` was being imported inside the item loop in `updatePlio()` inside `API/Plio.js` - fixed that
- [ ] Added one more sanity check - if video URL is empty in the creator, plio updating will stop and only resume when the url is not empty
- [ ] In `Plio.js`, in `createPlio()` function, rewritten the promise in a more efficient way
- [ ] Updated the fields in the object being passed to `PlioAPIService.updatePlio` -- turning JS variables into DB fields directly without the conversion that was happening before
- [ ] Removed `created_by` hardcoding from `Plio.js` API Service and moved it to `Editor.vue`
- [ ] Changed frontend/backend to work only with seconds - discarded milliseconds

src/pages/Editor.vue
deepansh96: Will explain on call

dalmia:
> * [x]  Removed the static endpoints from `.env` file and moved them to `src/services/API/Endpoints.js`
> * [x]  Returning the whole `video` object instead of just the `video.id` in `Video.js` - Vaibhav's second comment
> * [x]  Using `PUT` and not `PATCH` everywhere - will optimise with `PATCH` later - made an issue to track this #145
> * [x]  Keeping the unused old API endpoints - will be removed after aman's PR - created an issue for this - #146
> * [x]  Changed imported API function names according to Aman's suggestion. Eg - from `import PlioService` to `import PlioAPIService`
> * [x]  While filtering and retrieving the items for a particular plio, added an `order_by` on the backend serialiser to send the sorted by time list of items.
> * [x]  While creating a new question, earlier the values (two empty options, empty text, correct answer 0) was being hardcoded -- now it's returning from a function
> * [x]  `clonedeep` was being imported inside the item loop in `updatePlio()` inside `API/Plio.js` - fixed that
> * [x]  Added one more sanity check - if video URL is empty in the creator, plio updating will stop and only resume when the url is not empty
> * [x]  In `Plio.js`, in `createPlio()` function, rewritten the promise in a more efficient way
> * [x]  Updated the fields in the object being passed to `PlioAPIService.updatePlio` -- turning JS variables into DB fields directly without the conversion that was happening before
> * [x]  Removed `created_by` hardcoding from `Plio.js` API Service and moved it to `Editor.vue`
> * [x]  Changed frontend/backend to work only with seconds - discarded milliseconds

I think it would be better if the config file for endpoints stays at the same hierarchy as .env as one shouldn't have to go so dep to just add/edit/modify an endpoint - and it is like a config for the repo, right?

deepansh96:
> > * [x]   Removed the static endpoints from `.env` file and moved them to `src/services/API/Endpoints.js`
> > * [x]   Returning the whole `video` object instead of just the `video.id` in `Video.js` - Vaibhav's second comment
> > * [x]   Using `PUT` and not `PATCH` everywhere - will optimise with `PATCH` later - made an issue to track this #145
> > * [x]   Keeping the unused old API endpoints - will be removed after aman's PR - created an issue for this - #146
> > * [x]   Changed imported API function names according to Aman's suggestion. Eg - from `import PlioService` to `import PlioAPIService`
> > * [x]   While filtering and retrieving the items for a particular plio, added an `order_by` on the backend serialiser to send the sorted by time list of items.
> > * [x]   While creating a new question, earlier the values (two empty options, empty text, correct answer 0) was being hardcoded -- now it's returning from a function
> > * [x]   `clonedeep` was being imported inside the item loop in `updatePlio()` inside `API/Plio.js` - fixed that
> > * [x]   Added one more sanity check - if video URL is empty in the creator, plio updating will stop and only resume when the url is not empty
> > * [x]   In `Plio.js`, in `createPlio()` function, rewritten the promise in a more efficient way
> > * [x]   Updated the fields in the object being passed to `PlioAPIService.updatePlio` -- turning JS variables into DB fields directly without the conversion that was happening before
> > * [x]   Removed `created_by` hardcoding from `Plio.js` API Service and moved it to `Editor.vue`
> > * [x]   Changed frontend/backend to work only with seconds - discarded milliseconds
> 
> I think it would be better if the config file for endpoints stays at the same hierarchy as .env as one shouldn't have to go so dep to just add/edit/modify an endpoint - and it is like a config for the repo, right?

Again an opinion
The point of those endpoints being moved to a separate file was that they will not be changed anytime soon, very rarely.
My impression was that we can keep the API things clubbed in one folder otherwise it will spill out. We don't have an application wide config right now. Again, this is an opinion
Maybe @rathorevaibhav can weigh in?

src/pages/Editor.vue
```diff
@@ -839,25 +850,28 @@ export default {
...
-            return QuestionService.createQuestion({
+            return QuestionAPIService.createQuestion({
               item: createdItem.id,
-              type: "mcq",
-              options: ["", ""],
-              correct_answer: "0",
+              text: this.getDetailsForNewQuestion().text,
+              type: this.getQuestionTypeForNewQuestion(),
+              options: this.getDetailsForNewQuestion().options,
+              correct_answer: this.getDetailsForNewQuestion().correct_answer,
```
dalmia, line 869: why call the function 4 times -> the keys are the same so just pass the value of that function to the API no?

src/pages/Editor.vue
```diff
@@ -643,6 +644,9 @@ export default {
         this.hasUnpublishedChanges = true;
         return;
       }
+      // don't save plio if video URL is empty
+      if (this.videoURL == "") return;
```
dalmia, line 648: use `isVideoIDValid` computed property

src/services/API/Plio.js
```diff
@@ -42,67 +41,49 @@ export default {
...
     var newPlioData = {
       created_by: 1,
     };
-    return apiClient().post(process.env.VUE_APP_BACKEND_PLIOS, newPlioData);
+    return apiClient().post(pliosEndpoint, newPlioData);
   },
 
-  async updatePlio(value, plioId) {
+  async updatePlio(plioValue, plioId) {
+    // var cloneDeep = require("lodash.clonedeep");
```
dalmia, line 56: can remove this?

src/services/API/Plio.js
```diff
@@ -42,67 +41,49 @@ export default {
...
+
     // handle items and questions being updated
-    value.items.forEach((item) => {
+    plioValue.items.forEach((item) => {
       // pass a deepclone and not the reference
-      var cloneDeep = require("lodash.clonedeep");
-      var itemClone = cloneDeep(item);
-      ItemService.updateItem(itemClone);
-      QuestionService.updateQuestion(itemClone.details);
+      // var itemClone = cloneDeep(item);
```
dalmia, line 61: can remove this?

dalmia (changes requested):
Requested some minor changes - looks very good overall! :D

dalmia:
tested everything - looks great - this is great :D

dalmia (approved):

PR #150, opened 2021-04-08
Items overlap bug fix, css fixes on a smaller screen, dropdown rewrite, error states for time inputbox

deepansh96:
## Summary
- [x] Discarded the PrimeVue's dropdown component and written an implementation from scratch
- [x] Fixed a bug where the self written dropdown menu would cut into half
- [x] When in mobile mode - width < 640px - the `EDIT QUESTION` text that used to come on the item editor would disappear and the question dropdown would snap to the left most end
- [x] Bug fix - if an item marker was dragged on top of another item's marker, or dragged within 2 second range of another item's marker, it would not show a warning - after the fix - the marker would return to it's starting position in this case and a warning would be shown
- [x] Bug fix - when updating the time of an item from the time inputbox, if that time conflicted with the time of some other marker, no error or warning would be shown - this is now fixed
- [x] Publish button will be disabled now if there is any error at the time inputbox
- [x] Plio updation will stop now if there is any error at the time inputbox
- [x] Implemented configurable error states in the time inputbox. error states can be passed to the time inputbox component from the parent, and the component itself has some local error states. These error states are propagated up through the parent components to enable/disable the publish button
- [x] Refactored the function which found out if a timestamp was valid - being used at multiple places

## Test Plan
Tested locally


src/components/Editor/ItemEditor.vue
```diff
@@ -302,9 +313,31 @@ export default {
...
     },
+    isTimestampValid(timestamp, itemIndex = null) {
+      // loop through time values of items to check if the user selected time
+      // is valid or not
+      // using for loop instead of forEach as forEach was running async
+      for (let index = 0; index < this.localItemList.length; index++) {
+        // don't check against the selected item itself
+        if (index == itemIndex || itemIndex == null) continue;
+        var val = this.localItemList[index].time;
+        if (val == timestamp ||
```
dalmia, line 324: == wala logic is not needed as mentioned in another comment too!

dalmia (changes requested):

dalmia:
Creator - plio title not getting uploaded

src/pages/Editor.vue
```diff
@@ -521,8 +529,19 @@ export default {
...
+      var timeBeforeDragEnded = this.items[itemIndex].time;
       var itemTimestamp = this.itemTimestamps[itemIndex];
-      this.items[itemIndex]["time"] = itemTimestamp;
+
+      // check if the time after drag is valid and if not, set the item time
+      // back to the one before the drag
+      // else proceed with the new time
+      if (!this.isTimestampValid(itemTimestamp, itemIndex)) {
+        this.items[itemIndex]["time"] = timeBeforeDragEnded;
+        itemTimestamp = timeBeforeDragEnded;
```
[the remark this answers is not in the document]
deepansh96: This is needed here.
If we remove this statement and just return after showing the dialog, the marker will not move back to it's "before" position automatically. The dialog will show up yes, but the marker will stay where it was dragged to.

This is because when we drag the marker, `itemTimestamps` has already been updated with the new time position values. That is why I'm storing the before drag ended waala value in a separate variable

And we cannot also return from the function, we need the logic that is present outside the if and else waala part, to reset all the data and computed properties according to `timeBeforeDragEnded`

src/pages/Editor.vue
```diff
@@ -792,22 +811,22 @@ export default {
...
+    isTimestampValid(timestamp, itemIndex = null) {
       // loop through itemTimestamps to check if the time where the user
-      // is trying to add an item is valid or not
+      // is trying to add/update an item is valid or not
       // using for loop instead of forEach as forEach was running async
 
-      // returning -1 to signify that the time chosen by the user is not valid
       for (let index = 0; index < this.itemTimestamps.length; index++) {
+        // don't check against the item itself
+        if (index == itemIndex || itemIndex == null) continue;
```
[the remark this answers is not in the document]
deepansh96: So there can be two usecases of using this function - 
- using this function when adding a new item. In that case, we just want to check that `this.currentItemIndex` (which is where the user is trying to add a new item), is valid or not. For that, we just iterate through all times and return if its valid or not. In that case `itemIndex` argument of the function is not required
- using this function when updating the time of an already existing item. Let's say the item we're updating is called "X". In that case, we want to check that the new time of "X" should not conflict with all the times of all other items. BUT we don't want to check the new time of "X" with the old time of "X" itself, hence we also pass one `itemIndex`. Whenever we see that while iterating, we've hit "X" itself, we just `continue`

To answer your question - `itemIndex` will be null when this function is used to check a valid time for a new item

src/services/Functional/Item.js
```diff
@@ -1,3 +1,7 @@
+// what should the minimum time difference be
```
dalmia, line 1:
```suggestion
// what should the minimum time difference
```

src/services/Functional/Item.js
```diff
@@ -36,4 +40,26 @@ export default {
...
   },
+
+  isTimestampValid(timestamp, itemTimestamps, itemIndex = null) {
+    // loop through itemTimestamps to check if the time where the user
+    // is trying to add/update an item is valid or not
+    // using for loop instead of forEach as forEach was running async
+
+    for (let index = 0; index < itemTimestamps.length; index++) {
+      // don't check against the item itself
+      if (itemIndex == null || index == itemIndex) continue;
```
dalmia, line 51: logic fix

deepansh96, line 51:
```suggestion
      if (itemIndex != null && index == itemIndex) continue;
```

dalmia:
2 bugs:
- when inputting timestamp for item through input field, if timestamp is not valid - marker doesn't move but the slider marker is somewhere else (last position)
- after the error above, if you drag the marker to a valid position - still error text shows up

src/components/Editor/ItemEditor.vue
```diff
@@ -434,27 +453,19 @@ export default {
       // this object contains four keys - 'hour', 'minute', 'second'
       // and 'millisecond' - all are type Number
       get() {
-        // convert seconds to timeObject
+        // convert seconds to timeObject and
+        // check for any time input errors and toggle the respective error flags
+
         var itemTime = this.localItemList[this.localSelectedItemIndex].time;
+        this.checkTimeInputErrors(itemTime);
         return this.convertSecondsToISOTime(itemTime || 0);
```
dalmia, line 461: why keep || 0 ?

src/components/UI/Text/TimeInput.vue
```diff
@@ -184,6 +177,16 @@ export default {
...
+    timeObject: {
+      handler() {
+        // update the local values if the user selects a different item than the current selected one
+        this.localHour = this.timeObject.hour;
+        this.localMinute = this.timeObject.minute;
+        this.localSecond = this.timeObject.second;
+        this.localMillisecond = this.timeObject.millisecond;
+      },
+      deep: true,
+    },
```
dalmia, line 189: why are these properties not computed properties?

deepansh96: We want to store user ka locally entered value in these variables
If we make it computed toh everytime `timeObject` changes, they will be recalculated

dalmia (approved):

PR #152, opened 2021-04-09
API permissions

rathorevaibhav:
## Summary

1. Axios will append slash at the end of every API call. No need to define in the code now
2. Axios to set the access token and active organization in Header using interceptor
3. Active Workspace to be configured on frontend based on which API calls will be made
4. Active workspace is being updated in router/index.js before every call. Organization shortcode is determined from URL and cross-checked in vuex user.organizations array
5. Workspace switcher dropdown skeleton

## Test Plan

Nothing needed as of now.


src/components/UI/WorkspaceSwitcher.vue
```diff
@@ -0,0 +1,38 @@
...
+  </div>
+</template>
+
+<script>
+import { mapActions, mapState } from "vuex";
+
+export default {
+  name: "WorkspaceSwitcher",
+  computed: {
+    ...mapState('auth', ["activeWorkspace", "user"]),
```
dalmia, line 17: `user` is not being used anywhere.

src/router/index.js
```diff
@@ -61,13 +67,6 @@ const routes = [
     component: () => import("@/pages/Error"),
     props: { type: "404" },
   },
-  {
-    // Refer to: https://stackoverflow.com/a/64186073/7870587
-    path: "/:pathMatch(.*)*",
-    redirect: {
-      name: "404",
-    },
-  },
```
dalmia, line 70: why remove this? We kept this to redirect the user to 404 if they are visiting any URL not matching the routes above.

src/router/index.js
```diff
@@ -112,4 +111,25 @@ router.beforeEach((to, from, next) => {
...
 
+// set organization in vuex state if the route org parameter is in vuex user organizations array
+router.beforeEach((to, from, next) => {
+  if (
+    typeof to.params.org !== "undefined" &&
+    store.getters["auth/isAuthenticated"]
+  ) {
+    let orgShortcodes = store.state.auth.user.organizations.map(
+      (organization) => organization.shortcode
+    );
```
dalmia, line 122: If a user is added to an organization after they have already logged in, the `user.organizations` will still reflect the value from before the user was added to the org (right after they first logged in) right?

src/router/index.js
```diff
@@ -112,4 +111,25 @@ router.beforeEach((to, from, next) => {
...
+    store.dispatch("auth/setActiveWorkspace", to.params.org);
+    if (!orgShortcodes.includes(to.params.org)) {
+      // redirect to default page without organization id
+      store.dispatch("auth/setActiveWorkspace", to.params.org);
```
dalmia, line 126: shouldn't this be setting the workspace to Personal if want to redirect to default page "without organization id"?

dalmia (changes requested):

src/router/index.js
rathorevaibhav: Yeah, for that, we probably need to refresh the user data or do a reload after we create the UI for assigning users to an organization.

rathorevaibhav: Yeah, they'll have to re-login. Maybe we can do another API call after the organization assignment (from UI when we build it).

src/router/index.js
rathorevaibhav: sorry was experimenting. Putting it back.

src/components/UI/WorkspaceSwitcher.vue
rathorevaibhav: It is, right below in the workspaces computed property

src/components/UI/WorkspaceSwitcher.vue
```diff
@@ -0,0 +1,38 @@
+<template>
+  <div>
+    <select v-model="selectedWorkspace" @change="this.updateActiveWorkspace">
+      <option value="">Personal Workspace</option>
+      <option v-for="workspace in this.workspaces" :key="workspace.id" :value="workspace.shortcode">{{workspace.name}}</option>
+    </select>
+    <user-properties ref="userProperties"></user-properties>
```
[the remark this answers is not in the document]
rathorevaibhav: yeah. removed. Was a copy/paste thing

src/router/index.js
rathorevaibhav: yeah it should be unsetActiveWorkspace so that it default to empty string - which is the personal workspace

src/components/UI/WorkspaceSwitcher.vue
```diff
@@ -0,0 +1,35 @@
...
+  data() {
+    return {
+      selectedWorkspace: "",
+    }
+  },
+  methods: {
+    ...mapActions('auth', ['setActiveWorkspace']),
+    updateActiveWorkspace() {
+      this.setActiveWorkspace(this.selectedWorkspace);
+      this.$router.push({name: 'Home', params: { org: this.activeWorkspace }});
```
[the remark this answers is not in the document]
rathorevaibhav: The parent for this is `App.vue`. Maybe to avoid making it too big, we can have another file or maybe keep the redirect logic here only for now. What do you think?

I think once we have more pages and a better navigation in place, then we can figure it out. I was thinking more in terms of Notion that instead of dropdown, we should be showing all the organizations in a sidebar or something and user will click on those links. So they would see the organization home screen when they switch workspace.

src/components/UI/WorkspaceSwitcher.vue
dalmia: Sorry my bad!

src/router/index.js
dalmia: tracking it in #170

src/services/API/Plio.js
```diff
@@ -5,10 +5,15 @@ import QuestionAPIService from "@/services/API/Question.js";
 import { pliosEndpoint, itemsEndpoint } from "@/services/API/Endpoints.js";
 
 export default {
-  async getPlio(plioId) {
+  async getPlio(plioId, playMode = false) {
     // returns the details for one plio
```
dalmia, line 9: please just add one line for what does playMode mean (for future reference)

src/services/API/RootClient.js
```diff
@@ -28,6 +29,13 @@ client.interceptors.request.use(
       if (store.state.auth.accessToken) {
         config.headers.Authorization = `Bearer ${store.state.auth.accessToken.access_token}`;
       }
+
+      let currentRoute = Router.currentRoute.value;
+      if (currentRoute.params.org != undefined) {
+        store.dispatch("auth/setActiveWorkspace", currentRoute.params.org);
```
dalmia, line 35: if someone doesn't belong to an organization and if they are visiting `/af/play/PLIO_ID` this will set their workspace to AF right? Even when they don't belong to that workspace?

dalmia:
- bug when using workspace switcher

src/services/API/RootClient.js
rathorevaibhav: Yes, this will set it to that.

We will have the backend address the API permissioning part and return corresponding Response status.

dalmia (approved):

PR #180, opened 2021-04-19
Dockerization

### LINKED ISSUE #179: Dockerization
**Describe the task**
For the AWS deployment plan, we need to dockerize the current app.

**Expected behavior**
1. Make the app dockerized
2. Update installation instructions

rathorevaibhav:
Fixes #179 

## Summary

1. Removed `.env.production` and `.env.staging` to avoid environment-related data going into git. For every environment, we'll use `.env` by copying `.env.example`.
2. Dockerization for frontend
3. Updated instructions

## Test Plan

Nothing needed as of now.


.env.example
```diff
@@ -1,11 +1,21 @@
...
 VUE_APP_I18N_LOCALE=hi
 VUE_APP_I18N_FALLBACK_LOCALE=en
 
+# google oauth credentials
 VUE_APP_GOOGLE_CLIENT_ID=''
 
-VUE_APP_BACKEND_API_CLIENT_ID=''
-VUE_APP_BACKEND_API_CLIENT_SECRET=''
+# web server
+APP_PORT=8008
```
dalmia, line 21: this is meant to be 8080 right?

Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+COPY package*.json ./
+RUN npm install
+COPY . .
+RUN npm run build
+
+# production stage
+FROM nginx:stable-alpine as production-stage
+COPY --from=build-stage /app/dist /usr/share/nginx/html
+EXPOSE 80
+CMD ["nginx", "-g", "daemon off;"]
```
dalmia, line 13: should these be run based on whether we are running locally or remotely or as these to be run for each env? Sorry I didn't understand what does production stage mean here given the npm run build should ideally run vue right? Also, any reference for this?

docker-compose.yml
```diff
@@ -0,0 +1,13 @@
+version: '3.7'
+
+services: # the different images that will be running as containers
+  web: # service name
+    build: . #build the image for the web service from the dockerfile in parent directory.
```
dalmia, line 5:
```suggestion
    build: . # build the image for the web service from the dockerfile in parent directory.
```

docs/INSTALLATION.md
```diff
@@ -32,29 +34,23 @@ Follow the steps below to set up the repo for development
     git clone https://github.com/avantifellows/plio-frontend.git
     cd plio-frontend
     ```
-2. Install the dependencies
-    #### Development
+2. Make sure Docker is installed and docker version is running fine.
```
dalmia, line 37: same comment from the backend repo about a note on starting Docker Desktop when doing this locally.

dalmia (changes requested):

dalmia:
We also need to update `ivideo_frontend` in `package.json` and `package-lock.json` to `plio_frontend`

.env.example
rathorevaibhav: yeah my bad. was experimenting.

docs/INSTALLATION.md
```diff
@@ -34,7 +34,7 @@ Follow the steps below to set up the repo for development
     git clone https://github.com/avantifellows/plio-frontend.git
     cd plio-frontend
     ```
-2. Make sure Docker is installed and docker version is running fine.
+2. Make sure Docker Desktop application is running and docker version is giving a proper output.
```
dalmia, line 37: The Docker Desktop part is only true for setting it up locally right? Would be good to make that segregation.

Dockerfile
rathorevaibhav: I just updated it. We're using multi-stage feature in docker to have configurations set for all types of envs from single place. Earlier, people used to have different dockerfile for every environment.

For VueJS, I've used the official guide for VueJS2
https://vuejs.org/v2/cookbook/dockerize-vuejs-app.html

docker-compose.yml
```diff
@@ -2,12 +2,14 @@ version: '3.7'
...
+      context: . # build the image for the web service from the dockerfile in parent directory.
+      target: development-stage
     volumes:
       - .:/app # map data and files from parent directory in host to plio directory in docker container
     env_file: # file where env variables are stored. Used as best practice so as not to expose secret keys
       - .env # name of the env file
     image: plio-frontend # name of the image
     ports:
-      - "${APP_PORT}:80"
+      - "${APP_PORT}:${APP_PORT}"
```
dalmia, line 14: this was correct before right? as you were exposing port 80 in DockerFile?

docker-compose.yml
```diff
@@ -2,12 +2,14 @@ version: '3.7'
 
 services: # the different images that will be running as containers
   web: # service name
-    build: . # build the image for the web service from the dockerfile in parent directory.
+    build:
+      context: . # build the image for the web service from the dockerfile in parent directory.
+      target: development-stage
```
dalmia, line 7: the development stage is needed for running it locally, right?
Where are we setting the production stage as the target?

dalmia: also, would be great to have the reference added to Dockerfile: https://vuejs.org/v2/cookbook/dockerize-vuejs-app.html

docker-compose.yml
rathorevaibhav: Docker-compose will only be used for local development. So we're not really using nginx for local workspace and instead relying on `npm run serve` command.

docker-compose.yml
rathorevaibhav: Docker-compose only works for local setup.

Production by default will run Dockerfile only.

docker-compose.yml
dalmia: Got it - but where are we supplying the APP_PORT to be used for running `npm run serve` within the container? Once that is done, then mapping it to the port outside makes sense to me.

rathorevaibhav: oh yes, that's right. let me check.

rathorevaibhav: Done. Please check now.

dalmia (approved):

PR #184, opened 2021-04-20
Bug fixes and new progress bar implementation

### LINKED ISSUE #182: Home page table turns into a skeleton when "Create" button is clicked
**Describe the bug**
When clicking "create" on the home page, as the creator page is being loaded, `pending` becomes true and because of that the table skeleton shows up.

**Expected behavior**
After clicking "create" on the home page, the table should look as it is, no skeleton should be shown and the user should be routed to the creator

### LINKED ISSUE #183: After clicking logout, `this.user` is null and currently there's no null check for it - Home.vue & App.vue

deepansh96:
Fixes #182
Fixes #183 

## Summary

- [x] The table will NOT turn into a skeleton once "Create" button is clicked
- [x] Proper null checks have been added so when logging out, the console will not show an error when trying to find `status` inside `this.user`
- [x] Added an application level progress bar. It has been configured on `App.vue`. The progress bar will start and finish depending on the value of `this.pending` which comes via the vuex store. Benefit -- If for some process we don't want to toggle `this.pending` but we still want to show progress bar, we can skip doing `this.startLoading()` or `this.stopLoading()` and just use `this.$Progress.start()` or `this.$Progress.finish()`

## Test Plan
Tested locally

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -69,6 +69,7 @@
     <toast ref="toast"></toast>
     <router-view />
   </div>
+  <vue-progress-bar></vue-progress-bar>
 </template>
 
 <script>
@@ -110,6 +111,12 @@ export default {
     // set locale based on their config
     this.$refs.userConfig.setLocaleFromUserConfig();
   },
+  watch: {
+    pending(value) {
+      if (value) this.$Progress.start();
+      else this.$Progress.finish();
+    },
+  },
   methods: {
     // object spread operator
     // https://vuex.vuejs.org/guide/state.html#object-spread-operator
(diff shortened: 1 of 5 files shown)


src/App.vue
```diff
@@ -177,7 +185,7 @@ export default {
     },
     isUserApproved() {
       // whether the user is an approved user or in waitlist
-      return this.user.status == "approved";
+      return this.user != null && this.user.status == "approved" ? true : false;
```
dalmia, line 188:
```suggestion
      return this.user != null && this.user.status == "approved"
```

src/pages/Home.vue
```diff
@@ -72,7 +72,7 @@ export default {
     ...mapState("auth", ["activeWorkspace", "user"]),
     isUserApproved() {
       // whether the user is an approved user or in waitlist
-      return this.user.status == "approved";
+      return this.user != null && this.user.status == "approved" ? true : false;
```
dalmia, line 75:
```suggestion
      return this.user != null && this.user.status == "approved";
```

dalmia (approved):

PR #188, opened 2021-04-21
Error states

### LINKED ISSUE #106: Show appropriate outputs for different errors
For example, if there is an internet issue, say that cannot connect to the internet - does not say so. Similarly, handle other error codes gracefully.

deepansh96:
Fixes #106 

## Summary
- [x] Added a package [vue-toastification](https://github.com/Maronato/vue-toastification/tree/next) and replaced existing toast implementation with this
- [x] Added a toast which will pop up if the user enters a wrong OTP
- [x] For any errors that are caught while doing google auth, a toast saying "Something went wrong. Please try again" will popup
- [x] All responses returned by the API calls will be intercepted by an interceptor and network errors will be handled -- a toast saying "A network error occurred"
- [x] Added a generic file ErrorHandling.js in `services/Functional/` where the network error state handling is present. More types of generic error states can be specified there

## Test Plan
tested locally

### DIFF
--- a/src/components/UI/Text/URL.vue
+++ b/src/components/UI/Text/URL.vue
@@ -45,7 +45,6 @@ export default {
       type: [String, Object],
     },
   },
-  components: {},
   computed: {
     urlTextClass() {
       return [
--- a/src/components/UI/Text/URL.vue
+++ b/src/components/UI/Text/URL.vue
@@ -16,17 +16,16 @@
       >
       </inline-svg>
     </button>
-    <toast class="mt-20" ref="toast"></toast>
   </div>
 </template>
 
 <script>
-import Toast from "@/components/UI/Alert/Toast.vue";
+import { useToast } from "vue-toastification";
 
 export default {
   data() {
     return {
-      toastLife: 3000,
+      toast: useToast(), // use the toast component
     };
   },
   props: {
@@ -46,9 +45,7 @@ export default {
       type: [String, Object],
     },
   },
-  components: {
-    Toast,
-  },
+  components: {},
   computed: {
     urlTextClass() {
       return [
@@ -68,11 +65,8 @@ export default {
       hiddenElement.select();
       var success = document.execCommand("copy");
       document.body.removeChild(hiddenElement);
-      if (success) {
-        this.$refs.toast.show("success", "URL Copied Successfully", this.toastLife);
-      } else {
-        this.$refs.toast.show("error", "Error while copying", this.toastLife);
-      }
+      if (success) this.toast.success("URL Copied Successfully");
+      else this.toast.error("Error while copying");
 
       this.$emit("copied", success);
     },
--- a/src/pages/Login.vue
+++ b/src/pages/Login.vue
@@ -101,7 +101,7 @@ export default {
       invalidOtp: false, // whether the OTP is invalid
       toast: useToast(), // use the toast component
       wrongOTPWarning: "Incorrect OTP entered. Please try again!", // Warning to show when entered OTP is incorrect
-      somethingWentWrongWarning: "Something went wrong. Please try again!",
+      genericWarning: "Something went wrong. Please try again!",
     };
   },
   computed: {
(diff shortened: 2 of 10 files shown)


src/pages/Player.vue
```diff
@@ -179,6 +180,8 @@ export default {
       sessionDBId: null, // id for this session in the plio DB table
       retention: [], // array to store video retention value
       lastTimestampRetention: null, // last recorded timestamp in the retention array
+      toast: useToast(), // use the toast component
+      somethingWentWrongWarning: "Something went wrong. Please try again!",
```
dalmia, line 184:
```suggestion
      genericWarning: "Something went wrong. Please try again!",
```

src/components/UI/Text/URL.vue
```diff
@@ -46,9 +45,7 @@ export default {
       type: [String, Object],
     },
   },
-  components: {
-    Toast,
-  },
+  components: {},
```
dalmia, line 48:
```suggestion
```

src/services/Functional/ErrorHandling.js
```diff
@@ -0,0 +1,10 @@
+import { useToast } from "vue-toastification";
+const toast = useToast();
+
+export default {
+  handleAPIErrors(error) {
+    // TODO - Add more types of errors here
+    if (error.message == "Network Error")
+      toast.error("A network error occurred.");
```
dalmia, line 8:
```suggestion
      toast.error("Please check your internet connection");
```

src/services/Functional/ErrorHandling.js
```diff
@@ -0,0 +1,10 @@
+import { useToast } from "vue-toastification";
+const toast = useToast();
+
+export default {
+  handleAPIErrors(error) {
+    // TODO - Add more types of errors here
```
dalmia, line 6: don't think we need this here

src/services/Functional/ErrorHandling.js
```diff
@@ -0,0 +1,10 @@
...
+    if (error.message == "Network Error")
```
dalmia, line 7: can't we do it using response code? is the message always going to be this?

src/services/API/RootClient.js
```diff
@@ -34,6 +35,18 @@ client.interceptors.request.use(
...
+  }
+);
+
+// handle errors in responses for API requests
+client.interceptors.response.use(
+  (config) => {
+    return config;
+  },
+  (error) => {
+    ErrorHandling.handleAPIErrors(error);
```
dalmia, line 49: Are you sure both the types of errors should be handled by the same thing? The request hasn't happened only in the upper one - so, there won't be any API errors no? the error there is just any error in the interceptor code right?

src/pages/Login.vue
```diff
@@ -98,6 +99,9 @@ export default {
       requestedOtp: false, // whether the user has requested OTP once
       resentOtp: false, // whether the user has requested to resend OTP
       invalidOtp: false, // whether the OTP is invalid
+      toast: useToast(), // use the toast component
+      wrongOTPWarning: "Incorrect OTP entered. Please try again!", // Warning to show when entered OTP is incorrect
+      somethingWentWrongWarning: "Something went wrong. Please try again!",
```
dalmia, line 104:
```suggestion
      generic: "Something went wrong. Please try again!",

```

src/services/Functional/ErrorHandling.js
deepansh96: https://github.com/axios/axios/issues/383

src/services/Functional/ErrorHandling.js
```diff
@@ -0,0 +1,11 @@
+import { useToast } from "vue-toastification";
+const toast = useToast();
+
+export default {
+  handleAPIErrors(error) {
+    if (!error.response) {
+      // https://github.com/axios/axios/issues/383
```
deepansh96, line 7:
```suggestion
```

PR #207, opened 2021-04-23
FIX: <500px, a fullscreen overlay appears

### LINKED ISSUE #206: CSS fixes + fullscreen button overlay when screen size < 500px
player modal - need to force fullscreen when modal appears on phone
+
some cleanups - padding, button sizing, colors, css fixes

deepansh96:
Fixes #206


src/pages/Player.vue
```diff
@@ -257,6 +270,24 @@ export default {
...
+        value: this.fullscreenToggleText,
+        class: "text-white text-lg font-bold",
+      };
+    },
+    fullscreenToggleText() {
+      // text to show for the fullscreen button
+      if (this.isFullscreen) {
+        return "Exit Fullscreen";
+      }
+      return "Fullscreen";
```
dalmia, line 285: we don't need to toggle the text here right?

tailwind.config.js
```diff
@@ -18,6 +18,7 @@ module.exports = {
     }),
     screens: {
       xsm: "360px",
+      "breakpoint-500": "500px",
```
dalmia, line 21:
```suggestion
      "xmd": "500px",
```

src/components/Items/Question/Body.vue
```diff
@@ -1,17 +1,19 @@
...
     <!-- option container -->
     <div class="flex mx-4 md:mx-6 xl:mx-10">
       <ul class="w-full">
-        <li class="list-none">
+        <li class="list-none gap-1 flex flex-col">
           <div
             v-for="(option, optionIndex) in options"
             :key="optionIndex"
-            class="p-2 sm:p-4 mb-2 text-lg md:text-xl lg:text-2xl rounded-md"
+            class="border p-2 pl-4 text-lg md:text-xl lg:text-2xl rounded-md mr-5"
```
dalmia, line 16: margin is best to not be added on only one side right?
`mx-5` instead of `mr-5`?

src/pages/Player.vue
dalmia: Instead of Fullscreen - Go Fullscreen

dalmia:
Bug: after answering modal, if coming back to the video - still showing fullscreen button
Screenshot 2021-04-23 at 11 04 30 PM

dalmia:
There seem to be too many lines - the interface earlier seems cleaner - the one with less lines (even removing the line above the buttons could be good
Screenshot 2021-04-23 at 11 06 19 PM

tailwind.config.js
deepansh96: I think we should keep it as "breakpoint-500" or maybe "bp-500" or "bpt-500"
Sometimes I forget what the exact values of `md` and `lg` are and having one more `xmd` will be painful as in I'll have to go and check again and again in the config file.
Plus it also restricts us. If we `bp-500` we could also do `bp-550` or whatever value we like!

deepansh96:
> Bug: after answering modal, if coming back to the video - still showing fullscreen button
> Screenshot 2021-04-23 at 11 04 30 PM

I didn't understand the bug here

tailwind.config.js
deepansh96: And for that if we have to change the above `xsm` I really don't mind. It's just what I came up with at that time. Even that is a bit wrong. In tailwind, extra small is usually referred to as `xs` instead of `xsm`

dalmia: let's go with bp

dalmia:
TODO
- fullscreen button issue
- bp breakpoints

src/pages/Player.vue
```diff
@@ -4,7 +4,10 @@
     <video-skeleton v-if="!isVideoIdValid"></video-skeleton>
     <div v-else class="flex relative shadow-lg">
       <!-- fullscreen button overlay -->
-      <div class="z-50 absolute breakpoint-500:hidden w-full h-full bg-transparent">
+      <div
+        v-if="showItemModal"
+        class="z-50 absolute bp-50:hidden w-full h-full bg-transparent"
```
dalmia, line 9: bp-500?
have you tested it?

deepansh96: Lol arey
haan test toh kia tha fullscreen bug ke lie
Dont know kaise chal gaya

deepansh96: Oh, <500px mein I tested where it was working,
> 500px mein it was still showing which I didn't test :P

PR #209, opened 2021-04-26
NEW: open and connect to a users websocket, live sync the user object

### LINKED ISSUE #208: Sync any changes in the user's organizations or status (using websockets) with the server
- Adding a user to an organization or updating any user-org info through an API call should sync with the frontend instantly

- updating a user's status (approved/waitlist) through an API call should sync with the frontend vuex store instantly

deepansh96:
Fixes #208
Requires [this](https://github.com/avantifellows/plio-backend/pull/148) to be merged

## Summary
1. When the vuex store is initialized, it will call a plugin `createUserWebSocket`
2. The plugin will open a websocket connection to the server with a `ws://` URL that's configured on the server
3. When a message is received by the client from the server, the plugin will update the user state and the user's status state by dispatching a vuex action
4. If the websocket connections stops or throws an error, it will try to connect again in every 2 seconds
5. As the user's status is updated (approved or waitlist), the `Home.vue` component will try fetching the plios again to refresh the home page
6. As the user's organizations are updated, the `WorkspaceSwitcher` component will be populated with the new values/removed of the old values

## Test Plan
Tested locally


src/pages/Home.vue
```diff
@@ -53,6 +53,10 @@ export default {
     async activeWorkspace() {
       if (this.isUserApproved) await this.fetchAllPlioIds();
     },
+    isUserApproved(value) {
+      // fetch plios again if user approval status changes
+      if (value) this.fetchAllPlioIds();
+    },
```
dalmia, line 59: why has this been added?

src/store/plugins/usersWebSocketPlugin.js
```diff
@@ -0,0 +1,33 @@
+import { usersEndpoint } from "@/services/API/Endpoints.js";
+
+export default function createUsersWebSocket() {
+  return function connect(store) {
+    var UsersWebsocket = new WebSocket(
+      process.env.VUE_APP_BACKEND_WEBSOCKET + usersEndpoint
```
dalmia, line 6: no change to `.env.example`?

dalmia (commented):
Minor comments

src/pages/Home.vue
deepansh96: Eg- I'm on "waitlist". I log in -> I'm shown the waitlist waala warning message
API se the superuser approves me -> frontend mein `isUserApproved` becomes `true`. BUT `hasAnyPlios` is still false and the data hasn't been fetched because we don't fetch data when the user is unapproved. 
So as soon as the approval status changes, we need to fetch again.

And calling `this.fetchAllPlioIds` won't be too expensive  if not doing it for the first time. As we store all plios ka data in the vuex store, we can reuse that if it exists on the store.

src/store/plugins/usersWebSocketPlugin.js
```diff
@@ -0,0 +1,33 @@
...
+
+    UsersWebsocket.onopen = () => {
+      console.log("WebSocket is open now.");
+    };
+
+    UsersWebsocket.onmessage = (event) => {
+      let user = JSON.parse(event.data).user;
+      store.dispatch("auth/updateUserStatus", user.status, { root: true });
+      store.dispatch("auth/setUser", user, { root: true });
+      store.dispatch("sync/stopLoading", null, { root: true });
```
dalmia, line 17: I don't see any startloading being triggered

src/store/plugins/usersWebSocketPlugin.js
```diff
@@ -0,0 +1,33 @@
...
+    var UsersWebsocket = new WebSocket(
+      process.env.VUE_APP_BACKEND_WEBSOCKET + usersEndpoint
+    );
+
+    UsersWebsocket.onopen = () => {
+      console.log("WebSocket is open now.");
+    };
+
+    UsersWebsocket.onmessage = (event) => {
+      let user = JSON.parse(event.data).user;
```
dalmia, line 14: this assumes that the socket will always receive data corresponding to a user only

src/store/plugins/usersWebSocketPlugin.js
```diff
@@ -0,0 +1,33 @@
...
+      store.dispatch("auth/updateUserStatus", user.status, { root: true });
+      store.dispatch("auth/setUser", user, { root: true });
+      store.dispatch("sync/stopLoading", null, { root: true });
+    };
+
+    UsersWebsocket.onclose = () => {
+      console.log("Websocket has been closed. Reconnecting in 2 seconds");
+      setTimeout(() => {
+        connect(store);
+      }, 2000);
```
dalmia, line 24: Better to keep 2000 as a global variable at the top of the file for better readability.

PR #227, opened 2021-05-07
FIX: multitenacy bugs

### LINKED ISSUE #226: A user (not being part of an org), after viewing that org's plio, should not be navigated to the org's workspace when clicking home.
For example - if a user visits `http://localhost:8080/#/af/play/xpvvaprghl`(a plio part of the `af` workspace), they should be able to access it. 
But after watching the plio, if they click the home button, they are routed to `http://localhost:8080/#/af/home` which is unwanted behaviour as that user is not a part of `af`.
They should be redirected to `http://localhost:8080/#/home`

deepansh96:
Fixes #226 

## Summary
- Refactored the user config implementation. Earlier it was implemented as a vue component and the required methods were called using `this.$refs` but now it is re-written as a simple JS file and methods are exported from that file. Now those methods will be used as `UserConfigService.METHOD_NAME`.
- After watching an org's plio, and assuming the user does not belong to that org, the user will NOT be routed to the org's homepage when clicking home. They will be redirected to their personal workspace
- In case any API call returns a response 404, the user will be redirected to the 404 page.

## Test Plan
tested locally


src/services/API/RootClient.js
```diff
@@ -90,6 +91,11 @@ client.interceptors.response.use(
         .catch((error) => console.log(error));
     }
 
+    if (status === 404) {
+      router.replace({ name: "404" });
+      return Promise.reject(error);
+    }
+
```
dalmia, line 98: this should be part of the ErrorHandling thing made below right?

dalmia: And only the router part can be added there. The promise.reject is anyways happening at the end of the function

src/components/UI/LocaleSwitcher.vue
```diff
@@ -1,20 +1,21 @@
 <template>
   <div>
-    <select v-model="$i18n.locale" @change="$refs.userConfig.updateLocale">
+    <select v-model="$i18n.locale" @change="updateLocale">
```
dalmia, line 3: can't we do this?
```suggestion
    <select v-model="$i18n.locale" @change="UserConfigService .updateLocale">
```

src/services/Config/User.js
```diff
@@ -0,0 +1,57 @@
...
+        clearInterval(redirectId);
+      }
+    }, 10);
+  },
+
+  updateLocale() {
+    // update the locale to the local and remote configs
+    console.log("here");
+    console.log(i18n);
+    console.log(store);
```
dalmia, line 33:
```suggestion
```

src/services/Config/User.js
```diff
@@ -0,0 +1,57 @@
...
+    UserAPIService.updateUserConfig(store.state.auth.userId, userConfig)
+      .then((response) => {
+        if (response.status == 200) {
+          console.log("Config updated successfully");
+        } else {
+          console.log("Error with config update");
+        }
+      })
+      .then(() => store.dispatch("auth/saveConfig", JSON.stringify(userConfig))) // update user config locally
+      .catch((err) => console.log(err));
```
dalmia, line 55: aren't we handling all errors in the interceptor now?

dalmia: I think in the interceptor we should log the error too.

src/services/Config/User.js
```diff
@@ -0,0 +1,57 @@
...
+import store from "@/store";
+import i18n from "@/services/Localisation/i18n.js";
+
+export default {
+  saveLocalUserConfig() {
+    // fetch user config and save locally
+    var userId = store.state.auth.userId;
+    UserAPIService.getUserConfig(userId).then((response) => {
+      store.dispatch("auth/saveConfig", JSON.stringify(response.data));
+      // this.saveConfig(JSON.stringify(response.data)); // save user config locally
```
dalmia, line 11:
```suggestion
      // save user config locally
```

src/services/Config/User.js
```diff
@@ -0,0 +1,57 @@
...
+
+  setLocaleFromUserConfig() {
+    // set the system locale from the user config
+    var redirectId = setInterval(() => {
+      var userConfig = store.state.auth.config;
+      if (userConfig != null) {
+        userConfig = JSON.parse(userConfig);
+        i18n.global.locale =
+          userConfig["locale"] || process.env.VUE_APP_I18N_LOCALE;
+        // this.$i18n.locale = userConfig["locale"] || process.env.VUE_APP_I18N_LOCALE;
```
dalmia, line 23:
```suggestion
```

src/App.vue
```diff
@@ -119,6 +117,18 @@ export default {
         this.logoutUser();
       }
     },
+    onHomePage(value) {
+      // when coming to homepage from some other page, it's possible that the current user
+      // does not belong to the active workspace that was set on the page before.
+      // This logic will check if the current user actually belongs to the activeWorkspace
+      // set in the store. If not, then redirect to personal workspace
```
dalmia, line 124: seems too verbose.
```suggestion
      // check if the current user actually belongs to the activeWorkspace
      // set in the store. If not, then redirect to the personal workspace
```

src/App.vue
```diff
@@ -119,6 +117,18 @@ export default {
...
+      if (value && this.isAuthenticated) {
```
dalmia, line 125: why is the authenticated check needed now?

src/App.vue
```diff
@@ -119,6 +117,18 @@ export default {
...
+        var isUserInWorkspace = this.user.organizations.some((org) => {
+          return org.shortcode == this.activeWorkspace;
+        });
+        if (!isUserInWorkspace) this.$router.replace({ name: "Home" });
```
dalmia, line 129: don't you also need to reset the active workspace to personal?

src/App.vue
```diff
@@ -119,6 +117,18 @@ export default {
         this.logoutUser();
       }
     },
+    onHomePage(value) {
```
dalmia, line 120:
```suggestion
    onHomePage(newValue) {
```

src/App.vue
dalmia, line 125:
```suggestion
      if (newValue) {
```

src/components/UI/LocaleSwitcher.vue
```diff
@@ -1,20 +1,21 @@
...
 <script>
-import UserConfig from "@/services/Config/User.vue";
+import UserConfigService from "@/services/Config/User.js";
 
 export default {
   name: "LocaleSwitcher",
-  components: {
-    UserConfig,
+  methods: {
+    updateLocale() {
```
dalmia, line 16: 1 line comment

PR #249, opened 2021-05-15
Refresh cubejs access token on expiry

### LINKED ISSUE #248: CubeJS auth token would expire in some time and needs to be refetched

dalmia:
Fixes #248 

## Summary
- checks if the time difference between current time of request and the time when access token was fetched has crossed the expiry time.

## Test Plan
Tested locally.

### DIFF
--- a/src/services/API/RootClient.js
+++ b/src/services/API/RootClient.js
@@ -108,7 +108,7 @@ export function apiClient() {
 export function analyticsAPIClient() {
   return cubejs(
     async () => {
-      if (!store.getters["auth/isAccessTokenValid"]) {
+      if (!store.getters["auth/isAnalyticsAccessTokenValid"]) {
         await store.dispatch("auth/getAnalyticsAccessToken");
       }
       return store.state.auth.analyticsAccessToken;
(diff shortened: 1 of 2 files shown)


src/store/modules/auth.js
```diff
@@ -28,6 +30,15 @@ const getters = {
     }
     return activeOrganizationSchema;
   },
+  isAccessTokenValid: (state) => {
+    if (state.analyticsAccessToken === null) return false;
+    const currentTime = new Date();
+    const timeDifference =
+      (currentTime - state.analyticsAccessTokenFetchTime) / 1000;
```
deepansh96, line 37:
```suggestion
    const timeDifference =
      (new Date() - state.analyticsAccessTokenFetchTime) / 1000;
```
This won't work?

src/services/API/RootClient.js
```diff
@@ -108,7 +108,7 @@ export function apiClient() {
 export function analyticsAPIClient() {
   return cubejs(
     async () => {
-      if (store.state.auth.analyticsAccessToken === null) {
+      if (!store.getters["auth/isAccessTokenValid"]) {
```
deepansh96, line 111: better to change the getter to `isAnalyticsAccessTokenValid`

PR #260, opened 2021-05-17
FIX: ios visual bugs + modal bug

### LINKED ISSUE #261: iOS Bugs
- `gap` doesn't work with `flex`. Elements appear squished together
- The default `select` box's styling doesn't go along with other elements on the platform (on safari)
- Modal is not popping up when fullscreen
- [NON iOS] The text boxes for options -- Delete button visible over the text. Add some space between them

deepansh96:
Fixes #261 

## Test Plan
Tested locally and on staging - both non ios and ios browsers


src/pages/Player.vue
```diff
@@ -5,7 +5,7 @@
     <div v-else class="flex relative shadow-lg">
       <!-- fullscreen button overlay -->
       <div
-        v-if="showItemModal"
+        v-if="showItemModal && !isFullscreen"
```
dalmia, line 8: why this?

PR #266, opened 2021-05-18
UI bugs 0.1.3

### LINKED ISSUE #265: UI Bugs for release 0.1.3
https://www.notion.so/avantifellows/v0-1-3-UI-bugs-06bbe19217f34a4fbebe5c4647b3ebdd

deepansh96:
Fixes #265

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -50,7 +50,7 @@
               {{ $t("nav.login") }}
             </button>
           </router-link>
-          <a href="#" v-if="isAuthenticated" @click="logoutUser(true)">
+          <a href="#" v-if="isAuthenticated" @click="logoutButtonClicked">
             <button
               class="bg-white-500 hover:text-red-500 text-black font-bold border-0 object-contain px-1 py-2"
             >
@@ -92,6 +92,8 @@ export default {
     };
   },
   async created() {
+    // reset the value of pending while creating the component
+    if (this.pending) this.stopLoading();
     // place a listener for the event of closing of the browser
     window.addEventListener("beforeunload", this.onClose);
     if (this.isAuthenticated) await this.fetchAndUpdateUser();
@@ -137,25 +139,25 @@ export default {
         if (!isUserInWorkspace) this.$router.replace({ name: "Home" });
       }
     },
-    $route(to, from) {
-      // show user logged out toast when app is routed to login page
-      // and only when the logout action was not triggered by the user
-      if (to.name == "Login" && from.name != undefined && !this.userClickedLogout) {
-        this.toast.error(this.$t("error.auto_logout"));
-      }
-    },
   },
   methods: {
     // object spread operator
     // https://vuex.vuejs.org/guide/state.html#object-spread-operator
     ...mapActions("auth", ["unsetAccessToken", "fetchAndUpdateUser"]),
     ...mapActions("sync", ["stopLoading"]),
-    logoutUser(isUserTriggered = false) {
+    logoutButtonClicked() {
       // set whether the logout action as triggered by the user or not
-      this.userClickedLogout = isUserTriggered;
+      this.userClickedLogout = true;
+      // logout the user
+      this.logoutUser();
+    },
+    logoutUser() {
       // logs out the user
       this.unsetAccessToken().then(() => {
-        this.$router.replace({ name: "Login" });
+        this.$router.replace({
+          name: "Login",
+          params: { userClickedLogout: this.userClickedLogout },
+        });
         // added here so that if someone clicks on logout while
         // some activity is pending
         this.stopLoading();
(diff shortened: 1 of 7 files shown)


src/components/UI/Navigation/Paginator.vue
```diff
@@ -70,6 +70,7 @@
             :key="pageNumber"
             :class="setPageButtonClass(pageNumber)"
             @click="setAndRouteToPage(pageNumber)"
+            :disabled="paginatorDetails.currentPage === pageNumber"
```
dalmia, line 73: disable first/previous/next/last buttons like this too.

src/App.vue
```diff
@@ -130,13 +137,22 @@ export default {
         if (!isUserInWorkspace) this.$router.replace({ name: "Home" });
       }
     },
+    $route(to, from) {
+      // show user logged out toast when app is routed to login page
+      // and only when the logout action was not triggered by the user
+      if (to.name == "Login" && from.name != undefined && !this.userClickedLogout) {
+        this.toast.error(this.$t("error.auto_logout"));
+      }
+    },
```
dalmia, line 146: better to include it in router/index.js

src/App.vue
```diff
@@ -130,13 +137,22 @@ export default {
...
   },
   methods: {
     // object spread operator
     // https://vuex.vuejs.org/guide/state.html#object-spread-operator
     ...mapActions("auth", ["unsetAccessToken", "fetchAndUpdateUser"]),
     ...mapActions("sync", ["stopLoading"]),
-    logoutUser() {
+    logoutUser(isUserTriggered = false) {
+      // set whether the logout action as triggered by the user or not
+      this.userClickedLogout = isUserTriggered;
```
dalmia, line 155: can first have separate function - clickLogout triggered by onclick where this param is set and then this function does not need to accept any param.

src/pages/Login.vue
```diff
@@ -287,6 +287,7 @@ export default {
             this.invalidOtp = true;
             this.toast.error(this.$t("login.otp.incorrect"));
```
dalmia, line 288: 1. both submit otp and google sign in button show loading spinner
2. while loading spinner, if someone leaves, pending is still true and so, spinner keeps running

src/router/index.js
```diff
@@ -102,6 +103,16 @@ router.beforeEach((to, from, next) => {
   // clear all toasts whenever the route changes
   toast.clear();
 
+  // show auto logout toast
+  if (
+    to.name == "Login" &&
+    from.name != undefined &&
+    from.name != "Login" &&
+    to.params["userClickedLogout"] == "false"
```
dalmia, line 111: would be good to instead use this so that it doesn't break when `userClickedLogout` is not passed some time in the future (as we might forget this then):
```python
to.params.get("userClickedLogout")
```

deepansh96: Javascript doesn't have a `get` method
And I checked with not passing `userClickedLogout`, as it's a string,`to.params["userClickedLogout"] == "false"` just gives the output as a boolean `false` so it won't break

src/pages/Login.vue
```diff
@@ -33,32 +33,22 @@
...
+        :disabled="!isSubmitOTPEnabled || submitOTPIconConfig.enabled"
+      ></icon-button>
       <!-- button to request resending OTP -->
       <icon-button
         @click="resendOtp"
         :titleConfig="resendOTPTitleConfig"
         :buttonClass="resendOTPButtonClass"
         class="mt-2"
-        :isDisabled="pending"
+        :isDisabled="pending || submitOTPIconConfig.enabled"
```
dalmia, line 51: pending is not needed here right? - had added during the same PR when adding spinner to submit OTP

src/pages/Login.vue
```diff
@@ -275,7 +270,7 @@ export default {
     },
     phoneLogin() {
       // invoked for logging in with Phone
-      this.startLoading();
+      this.submitOTPIconConfig.enabled = true;
```
dalmia, line 273: i think it is better to have a data property: `isSubmitOTPInProgress` and then, set `submitOTPIconConfig.enabled` = `isSubmitOTPInProgress`. Code makes more sense then.

src/pages/Login.vue
```diff
@@ -144,16 +144,19 @@ export default {
       toast: useToast(), // use the toast component
       warningIcon: require("@/assets/images/exclamation-circle-solid.svg"),
       isGoogleAuthDisabled: true, // whether the google auth button is disabled
-      submitOTPIconConfig: {
-        // config for the loading icon on the submit otp button
-        enabled: false,
-        iconName: "spinner-solid",
-        iconClass: "animate-spin h-4 object-scale-down text-white",
-      },
+      isSubmitOTPInProgress: false, // whether submit OTP button is in loading mode or not
```
dalmia, line 147:
```suggestion
      isSubmitOTPInProgress: false, // whether the OTP has been submitted and is being verified

```

PR #268, opened 2021-05-19
Bunches cubeJS queries on the home page

### LINKED ISSUE #240: Bunch cubeJS queries on the home page

dalmia:
Fixes #240 

## Summary
- instead of making multiple queries for each plio, makes one query to CubeJS for fetching the list of all unique users.

## Test Plan
- [x] Tested locally
- [ ] Tested on staging
- [ ] Tested on production

### DIFF
--- a/src/pages/Home.vue
+++ b/src/pages/Home.vue
@@ -99,6 +99,7 @@ export default {
     return {
       tableColumns: ["name", "number_of_viewers"], // columns for the table
       tableData: [],
+      uniqueUsersList: [] // holds the number of unique users for all the plios fetched
       // dummy table data - to show skeletons when data is being loaded
       dummyTableData: Array(5).fill({
         name: { type: "component", value: "" },
@@ -178,7 +179,9 @@ export default {
           this.numberOfPliosPerPage = response.data.page_size; // set the page size
           this.prepareTableData(response.data.results); // prepare the data for the table
         }
-      );
+      ).then(() => {
+
+      });
     },
 
     createNewPlio() {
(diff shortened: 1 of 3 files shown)


src/services/API/Plio.js
```diff
@@ -130,6 +131,30 @@ export default {
...
+      resultSet.series()[0].series.forEach((seriesItem) => {
+        resultsMap[seriesItem.x] = seriesItem.value;
+      });
+
+    // holds the final list of values to be returned
+    var results = [];
+    plioIds.forEach((plioId) => {
+      // plios which do not have any sessions do not show up in
+      // the resultMap - use a default value for those plios
+      if (resultsMap[plioId] == undefined) results.push(0);
```
deepansh96, line 151:
```suggestion
      if (!(plioId in resultsMap)) results.push(0);
```
Seems more readable to me
There's no `not in` in javascript otherwise that would've been best

Reading the current code, it feels like its saying plioId key is present in resultsMap but the value of the key is undefined but actually plioId key doesn't exist in resultsMap.

I tested with this changes and it's working fine

src/pages/Home.vue
```diff
@@ -99,6 +99,7 @@ export default {
     return {
       tableColumns: ["name", "number_of_viewers"], // columns for the table
       tableData: [],
+      uniqueUsersList: [], // holds the number of unique users for all the plios fetched
```
deepansh96, line 102: Can change the name a little?
Feels like this array contains the unique users, not the number of unique users
Maybe `numberOfUniqueUsersList` or `countUniqueUsersList`
And i think we can skip the `list` waala part in the name

deepansh96 (approved):

PR #284, opened 2021-05-24
Access token expiration calculation had a bug

### LINKED ISSUE #283: Users are being logged out after 1 day

dalmia:
Fixes #283 

## Summary
- CubeJS token was getting expired but still CubeJS calls were being made which logged the user out
- CubeJS calls should not have been made with expired token but the code for checking if access token has expired had a bug

## Test Plan
- [x] Tested locally
- [ ] Tested on staging
- [ ] Tested on production

### DIFF
--- a/src/store/modules/auth.js
+++ b/src/store/modules/auth.js
@@ -32,7 +32,7 @@ const getters = {
   },
   isAnalyticsAccessTokenValid: (state) => {
     if (state.analyticsAccessToken === null) return false;
-    var currentTimeString = new Date().toString();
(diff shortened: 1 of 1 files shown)


src/store/modules/auth.js
```diff
@@ -32,8 +32,11 @@ const getters = {
   },
   isAnalyticsAccessTokenValid: (state) => {
     if (state.analyticsAccessToken === null) return false;
+    var currentTimeString = new Date(2021, 4, 26).toString();
```
dalmia, line 35: this is for testing on staging. Will remove upon testing

PR #293, opened 2021-05-26
Removed sorting for "name" column, fixed sorting for "number of viewers" column

### LINKED ISSUE #291: Sorting in home page table is incorrect
After the pagination implementation, the sorting feature (clicking the header of a column will sort the plios according to that column) was breaking and showing incorrect results.

This is because the logic was only sorting for the results already showing in the table and not sorting across all the plios.

FIX: 
- Remove sorting for "name" column for now
- Fix the sorting for "number of viewers" column
- To get the appropriate paginated response from the backend, we will need to hand over the ordering logic to the backend. This will mean that the backend will have to calculate the "unique number of viewers" and return the sorted results in a paginated format. Can not and should not use CubeJS on the backend

deepansh96:
Fixes #291 
Please see the issue linked for more details

## Summary
- user only sees the sorting arrow icon over "number of viewers" and not over "name" column
- user clicks the sorting button and a request to the backend is made asking for the ordered results in a paginated format

## Test Plan

- [x] Responsive
- [x] Tested locally
- [x] Tested on staging
- [ ] Tested on production

### DIFF
--- a/src/components/Collections/Table/Table.vue
+++ b/src/components/Collections/Table/Table.vue
@@ -203,7 +203,7 @@ export default {
       searchInputBoxClass:
         "w-full text-gray-700 leading-tight p-2 pl-4 focus:outline-none",
       // sort order for the "number of viewers" column. 1 - ascending, -1 - descending
-      numberOfViewersSortOrder: 1,
+      numViewersSortOrder: 1,
     };
   },
 
@@ -346,10 +346,13 @@ export default {
     },
     sortBy(columnName) {
       // toggle the sort order for "number_of_viewers" column
-      // and emit it to the parent (Home.vue)
+      // and emit the field to sort by, to the parent (Home.vue)
       if (columnName == "number_of_viewers") {
-        this.numberOfViewersSortOrder = this.numberOfViewersSortOrder * -1;
-        this.$emit("sort-by-number-of-viewers", this.numberOfViewersSortOrder);
+        this.numViewersSortOrder = this.numViewersSortOrder * -1;
+        this.$emit(
+          "sort-num-viewers",
+          this.numViewersSortOrder == -1 ? "-unique_viewers" : "unique_viewers"
+        );
       }
     },
     savePlioDetails(rowIndex, plioDetails) {
@@ -374,7 +377,7 @@ export default {
       // flip the icon if sort order changes
       // hide the icon for the "name" column
       return {
-        "transform rotate-180": this.numberOfViewersSortOrder == -1,
+        "transform rotate-180": this.numViewersSortOrder == -1,
         hidden: columnName == "name",
       };
     },
@@ -384,6 +387,6 @@ export default {
     },
   },
 
-  emits: ["search-plios", "reset-search-string", "sort-by-number-of-viewers"],
+  emits: ["search-plios", "reset-search-string", "sort-num-viewers"],
 };
 </script>
--- a/src/components/Collections/Table/Table.vue
+++ b/src/components/Collections/Table/Table.vue
@@ -63,24 +63,14 @@
                 <tr>
                   <th
                     v-for="(columnName, columnIndex) in columns"
-                    @click="sortBy(columnName)"
                     :key="columnName"
                     scope="col"
                     class="sm:py-3 py-1.5 text-left text-xs sm:text-md font-medium text-gray-500 uppercase tracking-wider w-2/3"
                     :class="getColumnHeaderStyleClass(columnIndex)"
                   >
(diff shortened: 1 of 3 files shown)


src/components/Collections/Table/Table.vue
```diff
@@ -344,64 +341,41 @@ export default {
...
     },
-    sortBy(key) {
-      // invoked when sorting arrows are clicked on a table column
-      // sortOrder value can be 1(ascending) or -1(descending)
-      // everytime this method is involed, sortOrder for the "key" is toggled
-      this.sortKey = key;
-      this.sortOrders[key] = this.sortOrders[key] * -1;
+    sortBy(columnName) {
+      // toggle the sort order for "number_of_viewers" column
+      // and emit it to the parent (Home.vue)
```
dalmia, line 349:
```suggestion
      // and emit it to the parent 
```

src/components/Collections/Table/Table.vue
```diff
@@ -344,64 +341,41 @@ export default {
...
+      if (columnName == "number_of_viewers") {
+        this.numberOfViewersSortOrder = this.numberOfViewersSortOrder * -1;
+        this.$emit("sort-by-number-of-viewers", this.numberOfViewersSortOrder);
```
dalmia, line 352: seems too long to me and don't think that is needed.
```suggestion
        this.$emit("sort-num-viewers", this.numberOfViewersSortOrder);
```

src/components/Collections/Table/Table.vue
```diff
@@ -206,11 +202,12 @@ export default {
       // classes for search bar input box
       searchInputBoxClass:
         "w-full text-gray-700 leading-tight p-2 pl-4 focus:outline-none",
+      // sort order for the "number of viewers" column. 1 - ascending, -1 - descending
+      numberOfViewersSortOrder: 1,
```
dalmia, line 206: seems too too long to me (generally, Of, For like words are not present in variable names)

```suggestion
      numViewersSortOrder: 1,
```

src/pages/Home.vue
```diff
@@ -112,6 +113,8 @@ export default {
       totalNumberOfPlios: 0, // total number of plios for the user
       numberOfPliosPerPage: 5, // number of plios to show on one page (default: 5)
       searchString: "", // the search string to filter the plios on
+      sortByFields: [], // array containing the fields to sort the plios on
+      currentPageNumber: undefined, // holds the current page number
```
dalmia, line 117: I don't see any utility for this?

src/pages/Home.vue
```diff
@@ -142,6 +145,18 @@ export default {
...
+
+      // remove any existing value of "unique_viewers" field. New value will be pushed below
+      this.sortByFields = this.sortByFields.filter((value) => {
+        return !value.includes("unique_viewers");
+      });
+
+      // push the appropriate field name according to the sort order and make the API call
+      this.sortByFields.push(sortOrder == 1 ? "unique_viewers" : "-unique_viewers");
+      await this.fetchPlioIds();
+    },
```
dalmia, line 159: why not keep the function as `sortPlios` and take the param to sort by as input (e.g. `unique_viewers`, `-unique_viewers`)? won't have to do the `== 1` check too and keeps it generic.

src/pages/Home.vue
```diff
@@ -142,6 +145,18 @@ export default {
...
     ...mapActions("plioItems", ["purgeAllPlios"]),
     ...mapActions("sync", ["startLoading", "stopLoading"]),
+    async sortByNumberOfViewers(sortOrder) {
+      // invoked when the user clicks the sort icon next to "number of viewers" column
+
+      // remove any existing value of "unique_viewers" field. New value will be pushed below
+      this.sortByFields = this.sortByFields.filter((value) => {
+        return !value.includes("unique_viewers");
+      });
+
```
dalmia, line 155: at any point we are sorting by only one thing, no? Or are you saying successive sorting is also applied?

src/services/API/Plio.js
```diff
@@ -64,6 +69,8 @@ export default {
     // add search string query param
     if (searchString != undefined && searchString != "")
       queryParams["search"] = searchString;
+    // add sort by query param
+    if (sortBy.length != 0) queryParams["ordering"] = sortBy.join();
```
dalmia, line 73: does `.join()` automatically add commas?

src/pages/Home.vue
dalmia: no multiple sorting - remove lists.

PR #298, opened 2021-05-28
FIX: reworked the login page

### LINKED ISSUE #296: Rework the login page
- add a string "+91" in the beginning of the phone input textbox
- Add a message saying the user opts in for the T&C by logging in
- move google sign in section above phone otp section
- make google sign in section bigger - no need to add any text as it's already visible in the button
- no need of any text for phone login - just show the input textbox with a placeholder - no buttons should be visible until phone number entered and is valid
- remove the message " only indian numbers supported "

deepansh96:
Fixes #296 

## Test Plan

- [x] Responsive
- [x] Tested locally
- [ ] Tested on staging
- [ ] Tested on production


src/components/UI/Text/InputNumber.vue
```diff
@@ -32,9 +32,16 @@
         <inline-svg :src="startIconObj"></inline-svg>
       </div>
 
+      <!-- static text -->
+      <div v-if="isStaticTextEnabled" :class="[staticText.class, staticTextDefaultClass]">
+        <p class="self-center">
+          {{ staticText.text }}
+        </p>
+      </div>
```
dalmia, line 40: not sure if this should be added here? Feels too specific? might be better to be added separately wherever it is used?

src/App.vue
```diff
@@ -180,10 +180,13 @@ export default {
         })
         .catch(() => this.toast.error(this.$t("error.create_plio")));
     },
-    onClose(event) {
+    onClose() {
```
dalmia, line 183: why remove the `event` argument? `event` later on will become undefined then right?

src/assets/locales/en.js
```diff
@@ -146,6 +146,7 @@ export default {
         valid: "",
         invalid: "Mobile number is invalid",
       },
+      input_placeholder: "Mobile Number",
```
dalmia, line 149: Enter Mobile Number

src/assets/locales/hi.js
```diff
@@ -146,6 +146,7 @@ export default {
         valid: "",
         invalid: "फ़ोन नंबर अमान्य है",
       },
+      input_placeholder: "मोबाइल नंबर",
```
dalmia, line 149: accordingly change here.

src/assets/locales/en.js
dalmia, line 149: placeholder should also not be in the center - it should be in the left - like the way we have in Editor

PR #329, opened 2021-06-11
Feature/testing

### LINKED ISSUE #328: Set up unit testing
**Describe the task**
As the Plio system is getting stable, we need to start writing test cases to test code stability at all times and avoid unprecedented issues.

**Expected behavior**
Set up VueJS unit test cases.

rathorevaibhav:
Fixes #328

## Summary

- CI workflow to have both pre-commit and test cases
- Test cases added for a few components
- packages installed to setup testing

## Test Plan

- [x] Tested locally
- [ ] Tested on staging
- [ ] Tested on production


.github/workflows/ci.yml
```diff
@@ -0,0 +1,31 @@
+name: Plio CI
```
dalmia, line 1: you might have to update the Github Workflows and Pre-commit doc file to point to this file right?

rathorevaibhav: Done.

jest.config.js
```diff
@@ -0,0 +1,11 @@
+// configurations for jest unit testing
+
+module.exports = {
+  preset: "@vue/cli-plugin-unit-jest",
+  transform: {
+    "^.+\\.vue$": "vue-jest",
+  },
+  collectCoverage: true,
+  collectCoverageFrom: ["src/**/*.{js,vue}", "!src/main.js"],
```
dalmia, line 9: does `**` include all possible sub-directories like `src/*/*/*/*.js` as well?

jest.config.js
```diff
@@ -0,0 +1,11 @@
+// configurations for jest unit testing
+
+module.exports = {
+  preset: "@vue/cli-plugin-unit-jest",
+  transform: {
+    "^.+\\.vue$": "vue-jest",
+  },
```
dalmia, line 7: can you explain this a bit?

package.json
```diff
@@ -67,11 +71,22 @@
...
+        "rules": {},
+        "overrides": [
+            {
+                "files": [
+                "**/__tests__/*.{j,t}s?(x)",
+                "**/tests/unit/**/*.spec.{j,t}s?(x)"
+                ],
+                "env": {
+                "jest": true
+                }
```
dalmia, line 83: can you explain this a bit?

tests/unit/UI/Alert/DialogBox.spec.js
```diff
@@ -0,0 +1,29 @@
+import { mount } from "@vue/test-utils";
```
dalmia, line 1: I think these should follow the same folder structure starting from `src` right? So, there should be a `components` folder within which the `UI` folder should be there, right?

tests/unit/UI/Alert/DialogBox.spec.js
```diff
@@ -0,0 +1,29 @@
...
+import DialogBox from "@/components/UI/Alert/DialogBox";
+
+describe("DialogBox.vue", () => {
```
dalmia, line 4: should we use the full path? `components/UI/Alert/DialogBox`?

tests/unit/UI/Alert/DialogBox.spec.js
```diff
@@ -0,0 +1,29 @@
...
+
+  it("renders description correctly", () => {
+    const description = "Dialog Title";
+    const wrapper = mount(DialogBox, {
+      props: {
+        description,
+      },
+    });
+    expect(wrapper.find("p").text()).toBe(description);
+  });
```
dalmia, line 28: shouldn't we ideally run one more with both of them being passed to ensure that both the correct `p` tags are being set?

tests/unit/UI/Alert/Toast.spec.js
```diff
@@ -0,0 +1,9 @@
+import { mount } from "@vue/test-utils";
+import Toast from "@/components/UI/Alert/Toast";
+
+describe("Toast.vue", () => {
+  it("should render with default values", () => {
+    const wrapper = mount(Toast);
+    expect(wrapper).toBeTruthy();
```
dalmia, line 7: this just checks that it is not empty or something right?

tests/unit/UI/Alert/Toast.spec.js
```diff
@@ -0,0 +1,9 @@
+import { mount } from "@vue/test-utils";
```
dalmia, line 1: some of the same comments as the dialog box file

tests/unit/UI/Badges/SimpleBadge.spec.js
```diff
@@ -0,0 +1,26 @@
+import { mount } from "@vue/test-utils";
```
dalmia, line 1: some of the same comments as the dialog box file

tests/unit/UI/Badges/SimpleBadge.spec.js
```diff
@@ -0,0 +1,26 @@
...
+import SimpleBadge from "@/components/UI/Badges/SimpleBadge";
+
+describe("SimpleBadge.vue", () => {
+  it("renders title correctly", () => {
```
dalmia, line 5: if doing tests on things like title, description already covers the `truthy` part, we can remove the truthy checks from the other files too right?

Dockerfile
```diff
@@ -2,6 +2,7 @@
 FROM node:lts-alpine as base-stage
 WORKDIR /app
 COPY package*.json ./
+RUN apk --no-cache --virtual build-dependencies add python make g++
```
dalmia, line 5: can you please clarify why we need this?

jest.init.js
```diff
@@ -0,0 +1,12 @@
...
+ * It is being run from jest.config.js as a setupFile.
+ * It gets loaded before the tests are run.
+ */
+
+import { config } from "@vue/test-utils";
+import InlineSvg from "vue-inline-svg";
+
+config.global.components = {
+  InlineSvg,
+};
```
dalmia, line 12: I am maybe missing something but I don't see this being used anywhere in the test files - why do we need this? because the files we are testing are using this component and since they won't be referring to main.js that's why?

tests/unit/UI/DropDownMenu/Dropdown.spec.js
```diff
@@ -0,0 +1,30 @@
...
+
+  it("renders options list correctly", () => {
+    const wrapper = mount(Dropdown, {
+      props: {
+        optionsList,
+      },
+    });
+    expect(wrapper.find("option").html()).toContain(optionsList[0].label);
+    expect(wrapper.find("option").element.value).toMatch(optionsList[0].value);
+  });
```
dalmia, line 29: I think we should also just add the check for the second element too!

tests/unit/UI/Buttons/IconButton.spec.js
```diff
@@ -0,0 +1,16 @@
...
+  it("renders title correctly", () => {
+    const buttonText = "Test Button";
+    const wrapper = mount(IconButton, {
+      props: {
+        titleConfig: {
+          value: buttonText,
+        },
+      },
+    });
+    expect(wrapper.text()).toMatch(buttonText);
```
dalmia, line 14: where can we see the outputs of some of these things while developing - say, by using console.log here?

jest.config.js
rathorevaibhav: yep.

tests/unit/UI/Alert/DialogBox.spec.js
rathorevaibhav: aah okay. got it.

tests/unit/UI/Alert/DialogBox.spec.js
rathorevaibhav: Ideally, there's no harm. More the merrier!

I see that there can be more test cases added to this component and I think let's plan and work on them at once?

tests/unit/UI/Alert/Toast.spec.js
rathorevaibhav: Also, because if you set values, it overrides default values and those statements doesn't get executed. So code coverage decreases. So we have one test case at least when all of the default statements are executed.

tests/unit/UI/Badges/SimpleBadge.spec.js
rathorevaibhav: Not sure if I understood your question. Can you elaborate?

Dockerfile
rathorevaibhav: This is needed to run jest and other testing-related dependencies. I've added a comment.

tests/unit/UI/Buttons/IconButton.spec.js
rathorevaibhav: Yep. `console.log()` works. Here's an output of this line of code `console.log(wrapper.find('option').html());`

image

package.json
rathorevaibhav: These are the default configs that comes in with jest. This came in after I ran this command:
```
vue add unit-jest
```

jest.config.js
rathorevaibhav: I initially used a lot of articles to get the test cases working. I think this summarizes well.
https://vue-test-utils.vuejs.org/installation/#using-vue-test-utils-with-jest-recommended

jest.config.json got created automatically after running `vue add unit-jest`

dalmia (approved):

package.json
```diff
@@ -67,11 +71,22 @@
...
+            {
+                "files": [
+                    "**/__tests__/*.{j,t}s?(x)",
+                    "**/tests/unit/**/*.spec.{j,t}s?(x)"
+                ],
+                "env": {
+                    "jest": true
+                }
+            }
+        ]
```
deepansh96, line 85: This just means that eslint will skip these files right?

package.json
```diff
@@ -67,11 +71,22 @@
...
     },
     "browserslist": [
         "> 1%",
         "last 2 versions",
         "not dead"
     ]
-}
+    }
```
deepansh96, line 92: minor indentation change

deepansh96 (approved):

PR #352, opened 2021-06-21
NEW: initialised mixpanel, router for test cases + player initial testcase

### LINKED ISSUE #351: Initialise mixpanel + router globally for test cases

deepansh96:
Fixes #351 

## Test Plan

- [ ] Test Responsiveness
   - [ ] Laptop (1200px)
   - [ ] Tablet (760px)
   - [ ] Phone (320px)
- [ ] Cross-Browser Testing
   - [ ] Chrome
   - [ ] Firefox
   - [ ] Safari
- [ ] Local Language Support
- [x] Wrote tests
- [ ] Tested locally
- [ ] Tested on staging
- [ ] Tested on production

### DIFF
--- a/jest.init.js
+++ b/jest.init.js
@@ -6,7 +6,22 @@
 
 import { config } from "@vue/test-utils";
 import InlineSvg from "vue-inline-svg";
+import Toast from "vue-toastification";
 
-config.global.components = {
-  InlineSvg,
+import mixpanel from "mixpanel-browser";
+mixpanel.init(process.env.VUE_APP_MIXPANEL_PROJECT_TOKEN);
+const $mixpanel = mixpanel;
+
+import router from "@/router";
+const $router = router;
+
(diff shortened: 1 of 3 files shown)


jest.init.js
```diff
@@ -6,7 +6,22 @@
 
 import { config } from "@vue/test-utils";
 import InlineSvg from "vue-inline-svg";
+import Toast from "vue-toastification";
 
-config.global.components = {
-  InlineSvg,
+import mixpanel from "mixpanel-browser";
+mixpanel.init(process.env.VUE_APP_MIXPANEL_PROJECT_TOKEN);
```
dalmia, line 12: if you are using environment variables - you need to set them in `.github/workflows/ci.yml` but my understanding is that this is just a mock and hence, we don't need to initialize with our token and hence, this line can be removed/

PR #355, opened 2021-06-21
FIX: more tests

### LINKED ISSUE #354: More test cases for frontend

dalmia:
Fixes #354 

## Test Plan
- [ ] ~Test Responsiveness~
   - [ ] Laptop (1200px)
   - [ ] Tablet (760px)
   - [ ] Phone (320px)
- [ ] ~Cross-Browser Testing~
   - [ ] Chrome
   - [ ] Firefox
   - [ ] Safari
- [x] Wrote tests
- [x] Tested locally
- [ ] Tested on staging
- [ ] Tested on production

### DIFF
--- a/src/components/UI/Alert/DialogBox.vue
+++ b/src/components/UI/Alert/DialogBox.vue
@@ -38,6 +38,7 @@
             class="py-2 px-4 text-white w-full transition ease-in duration-200 text-center text-base font-semibold shadow-md rounded-lg"
             :class="confirmButtonClass"
             @click="confirmClicked"
+            data-test="confirmButton"
           >
             {{ confirmButtonText }}
           </button>
@@ -47,6 +48,7 @@
             class="py-2 px-4 w-full transition ease-in duration-200 text-center text-base font-semibold shadow-md rounded-lg"
             :class="cancelButtonClass"
             @click="cancelClicked"
+            data-test="cancelButton"
           >
             {{ cancelButtonText }}
           </button>
--- a/src/components/UI/Alert/DialogBox.vue
+++ b/src/components/UI/Alert/DialogBox.vue
@@ -10,11 +10,13 @@
           :src="icon"
           class="place-self-center"
           :class="iconClass"
+          data-test="icon"
         ></inline-svg>
         <!-- title -->
         <p
           class="text-gray-800 dark:text-gray-200 text-xl font-bold mt-4"
           v-if="showTitle"
+          data-test="title"
         >
           {{ title }}
         </p>
@@ -22,6 +24,7 @@
         <p
           class="text-gray-600 dark:text-gray-400 text-sm py-2 px-6"
           v-if="showDescription"
+          data-test="description"
         >
           {{ description }}
         </p>
--- a/tests/unit/components/Collections/ListItems/PlioListItem.spec.js
+++ b/tests/unit/components/Collections/ListItems/PlioListItem.spec.js
@@ -1,0 +1,31 @@
+import { mount } from "@vue/test-utils";
+import PlioListItem from "@/components/Collections/ListItems/PlioListItem";
+import Tooltip from "primevue/tooltip";
+
+// as window.matchMedia is not defined in the DOM
+Object.defineProperty(window, "matchMedia", {
+  writable: true,
+  value: jest.fn().mockImplementation((query) => ({
+    matches: false,
+    media: query,
+    onchange: null,
+    addListener: jest.fn(), // deprecated
+    removeListener: jest.fn(), // deprecated
+    addEventListener: jest.fn(),
+    removeEventListener: jest.fn(),
+    dispatchEvent: jest.fn(),
+  })),
+});
+
+describe("PlioListItem.vue", () => {
+  it("should render with default values", () => {
+    const wrapper = mount(PlioListItem, {
+      global: {
+        directives: {
+          tooltip: Tooltip,
+        },
+      },
+    });
+    expect(wrapper).toBeTruthy();
+  });
+});
--- a/tests/unit/components/Collections/ListItems/PlioListItem.spec.js
+++ b/tests/unit/components/Collections/ListItems/PlioListItem.spec.js
@@ -1,31 +1,9 @@
 import { mount } from "@vue/test-utils";
 import PlioListItem from "@/components/Collections/ListItems/PlioListItem";
-import Tooltip from "primevue/tooltip";
-
-// as window.matchMedia is not defined in the DOM
-Object.defineProperty(window, "matchMedia", {
-  writable: true,
-  value: jest.fn().mockImplementation((query) => ({
-    matches: false,
-    media: query,
-    onchange: null,
-    addListener: jest.fn(), // deprecated
-    removeListener: jest.fn(), // deprecated
-    addEventListener: jest.fn(),
-    removeEventListener: jest.fn(),
-    dispatchEvent: jest.fn(),
-  })),
-});
 
 describe("PlioListItem.vue", () => {
   it("should render with default values", () => {
-    const wrapper = mount(PlioListItem, {
-      global: {
-        directives: {
-          tooltip: Tooltip,
-        },
-      },
-    });
+    const wrapper = mount(PlioListItem);
     expect(wrapper).toBeTruthy();
   });
 });
--- a/tests/unit/components/Editor/ItemEditor.spec.js
+++ b/tests/unit/components/Editor/ItemEditor.spec.js
@@ -1,0 +1,30 @@
+import { mount } from "@vue/test-utils";
+import ItemEditor from "@/components/Editor/ItemEditor";
+import Tooltip from "primevue/tooltip";
+import VueClickAway from "vue3-click-away";
+
+describe("ItemEditor.vue", () => {
+  it("should render with required values", () => {
+    const wrapper = mount(ItemEditor, {
+      global: {
+        directives: {
+          tooltip: Tooltip,
+          clickAway: VueClickAway,
+        },
+      },
+      props: {
+        itemList: [
+          {
+            type: "question",
+            details: {
+              text: "test",
+              type: "subjective",
+            },
+            time: 13,
+          },
+        ],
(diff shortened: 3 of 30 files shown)


tests/unit/components/UI/Text/InputText.spec.js
```diff
@@ -0,0 +1,16 @@
+import { mount } from "@vue/test-utils";
+import InputText from "@/components/UI/Text/InputText";
+import Tooltip from "primevue/tooltip";
+
+describe("InputText.vue", () => {
+  it("should render with default values", () => {
+    const wrapper = mount(InputText, {
+      global: {
+        directives: {
+          tooltip: Tooltip,
```
rathorevaibhav, line 10: Do you think we should have the global directives also mentioned like global components? In one file `jest.init.js` instead of loading with every mount?

tests/unit/components/UI/Alert/DialogBox.spec.js
```diff
@@ -11,19 +15,66 @@ describe("DialogBox.vue", () => {
...
+      },
+      global: {
+        stubs: {
+          InlineSvg: InlineSvg,
+        },
       },
     });
-    expect(wrapper.find("p").text()).toBe(description);
+    expect(wrapper.get('[data-test="icon"]').html()).toContain("img");
+    expect(wrapper.get('[data-test="icon"]').classes()).toEqual(
```
rathorevaibhav, line 52: just curious, how is it different from `.toContain()`?

tests/unit/components/Collections/ListItems/PlioListItem.spec.js
```diff
@@ -0,0 +1,31 @@
+import { mount } from "@vue/test-utils";
+import PlioListItem from "@/components/Collections/ListItems/PlioListItem";
+import Tooltip from "primevue/tooltip";
+
+// as window.matchMedia is not defined in the DOM
+Object.defineProperty(window, "matchMedia", {
```
rathorevaibhav, line 6: should this also be in `jest.init.js` like the other window property you've defined there?

tests/unit/components/UI/Alert/DialogBox.spec.js
deepansh96: I'm guessing to contain will pass even if it's not an exact match and toEqual will have to be an exact match

tests/unit/components/UI/Alert/DialogBox.spec.js
```diff
@@ -11,19 +15,66 @@ describe("DialogBox.vue", () => {
...
+      expect.arrayContaining(["w-12", "h-12"])
+    );
+  });
+
+  it("confirm button present", () => {
+    const wrapper = mount(DialogBox);
+    expect(wrapper.get('[data-test="confirmButton"]').text()).toBe("Yes");
```
deepansh96, line 59: Isn't it best to just check if the button element exists instead of checking the text inside the button? The button text might change in the future and also if checking for text, we have to make sure that all corresponding i18n strings also match

tests/unit/components/UI/Alert/DialogBox.spec.js
```diff
@@ -11,19 +15,66 @@ describe("DialogBox.vue", () => {
...
+  });
+
+  it("cancel button present", () => {
+    const wrapper = mount(DialogBox);
+    expect(wrapper.get('[data-test="cancelButton"]').text()).toBe("No");
```
deepansh96, line 64: Same comment as above

tests/unit/components/UI/Alert/DialogBox.spec.js
```diff
@@ -11,19 +15,66 @@ describe("DialogBox.vue", () => {
...
+  });
+
+  it("confirm click works", () => {
+    const wrapper = mount(DialogBox);
+    wrapper.get('[data-test="confirmButton"]').trigger("click");
```
deepansh96, line 69: Shouldn't we await for this trigger? Everywhere on the docs it's written that the best practice is to await for a trigger to complete

tests/unit/components/UI/Alert/DialogBox.spec.js
```diff
@@ -11,19 +15,66 @@ describe("DialogBox.vue", () => {
...
+
+    expect(wrapper.emitted()).toHaveProperty("confirm");
+  });
+
+  it("cancel click works", () => {
+    const wrapper = mount(DialogBox);
+    wrapper.get('[data-test="cancelButton"]').trigger("click");
```
deepansh96, line 76: Same await comment as above

tests/unit/components/UI/Alert/ImageUploaderDialog.spec.js
```diff
@@ -0,0 +1,29 @@
+import { mount } from "@vue/test-utils";
+import store from "@/store";
```
deepansh96, line 2: Why import store here? It's already imported globally right?

tests/unit/components/Editor/ItemEditor.spec.js
```diff
@@ -0,0 +1,30 @@
+import { mount } from "@vue/test-utils";
+import ItemEditor from "@/components/Editor/ItemEditor";
+import Tooltip from "primevue/tooltip";
+import VueClickAway from "vue3-click-away";
```
deepansh96, line 4: We're importing clickaway and tooltip multiple times in multiple component tests. Wouldn't it make sense to just import it globally in jest config?

tests/unit/components/UI/Text/InputText.spec.js
deepansh96: Agreed

tests/unit/components/UI/Text/URL.spec.js
```diff
@@ -0,0 +1,81 @@
...
+  it("link text class is set correctly", async () => {
+    const link = "testLink";
+    const linkClass = "w-12 h-12";
+    await wrapper.setProps({
+      link: link,
+      urlStyleClass: linkClass,
+    });
+
+    expect(wrapper.get('[data-test="link"]').classes()).toEqual(
+      expect.arrayContaining(["w-12", "h-12"])
```
deepansh96, line 41: Isn't it better to cross check this with the variable "linkClass" that you created? Just trying to avoid hardcoding

tests/unit/components/UI/Text/URL.spec.js
```diff
@@ -0,0 +1,81 @@
...
+
+  it("link text underlining works correctly", async () => {
+    const link = "testLink";
+    await wrapper.setProps({
+      link: link,
+      isUnderlined: true,
+    });
+
+    expect(wrapper.get('[data-test="link"]').classes()).toEqual(
+      expect.arrayContaining(["border-b-2", "border-yellow-500"])
```
deepansh96, line 53: This hardcoding might cause pain later on. Isn't it better to actually retrieve the underline class from the component, save it in a variable and then assert the tested . classes () with that variable? This will help in long term because we keep changing the CSS classes often. Everytime we'll need to change the test cases. We can access the computed properties from a wrapper and hence we should be able to access the class which is the underline class here

tests/unit/components/UI/Text/URL.spec.js
```diff
@@ -0,0 +1,81 @@
...
+    );
+  });
+
+  it("button click is working correctly", async () => {
+    const link = "testLink";
+    await wrapper.setProps({
+      link: link,
+    });
+
+    wrapper.get('[data-test="copyButton"]').trigger("click");
```
deepansh96, line 76: Await, as the comment above

src/components/UI/Alert/DialogBox.vue
```diff
@@ -10,18 +10,21 @@
           :src="icon"
           class="place-self-center"
           :class="iconClass"
+          data-test="icon"
         ></inline-svg>
         <!-- title -->
         <p
           class="text-gray-800 dark:text-gray-200 text-xl font-bold mt-4"
           v-if="showTitle"
+          data-test="title"
```
deepansh96, line 19: Did you check that multiple data-tests with the same name do not conflict in the main app?

tests/unit/components/UI/Alert/DialogBox.spec.js
```diff
@@ -11,19 +15,66 @@ describe("DialogBox.vue", () => {
...
+      global: {
+        stubs: {
+          InlineSvg: InlineSvg,
+        },
       },
     });
-    expect(wrapper.find("p").text()).toBe(description);
+    expect(wrapper.get('[data-test="icon"]').html()).toContain("img");
+    expect(wrapper.get('[data-test="icon"]').classes()).toEqual(
+      expect.arrayContaining(["w-12", "h-12"])
```
deepansh96, line 53: Can avoid hardcoding

tests/unit/components/UI/Navigation/Paginator.spec.js
```diff
@@ -0,0 +1,13 @@
+import { mount } from "@vue/test-utils";
+import Paginator from "@/components/UI/Navigation/Paginator";
+
+describe("Paginator.vue", () => {
+  it("should render with required values only", () => {
+    const wrapper = mount(Paginator, {
+      props: {
+        totalItems: 10,
```
deepansh96, line 8: Can store this in a variable. Will have to change at one place only if required otherwise will have to change at 2 places

deepansh96 (changes requested):

tests/unit/components/UI/Alert/DialogBox.spec.js
dalmia: `toContain` won't work when the returned value is an array. So, if the actual value was `['a', 'b', 'c']`, `.toContain('a')` won't work.

tests/unit/components/UI/Alert/ImageUploaderDialog.spec.js
dalmia: that is for using `this.$store`. this is for accessing actions and state. will try setting it globally too.

deepansh96 (approved):

PR #361, opened 2021-06-23
Sentry integration

rathorevaibhav:
Targets application-level monitoring for #353 

## Summary

1. Sentry installation
2. Used JS app based sentry installation as Sentry doesn't support Vue3 yet.
3. Documentation update

## Test Plan

- [ ] Test Responsiveness
   - [ ] Laptop (1200px)
   - [ ] Tablet (760px)
   - [ ] Phone (320px)
- [ ] Cross-Browser Testing
   - [ ] Chrome
   - [ ] Firefox
   - [ ] Safari
- [ ] Local Language Support
- [ ] Wrote tests
- [x] Tested locally
- [ ] Tested on staging
- [ ] Tested on production


src/main.js
```diff
@@ -55,6 +59,25 @@ const filterBeforeCreate = (toast, toasts) => {
...
+  process.env.VUE_APP_SENTRY_DSN
+) {
+  // Since Vue3 isn't officially supported yet by Sentry, we're using
+  // the JavaScript Sentry integration with @sentry/browser.
+  Sentry.init({
+    dsn: process.env.VUE_APP_SENTRY_DSN,
+    integrations: [new Integrations.BrowserTracing()],
+    // Set tracesSampleRate to 1.0 to capture 100%
+    // of transactions for performance monitoring.
+    tracesSampleRate: 1.0,
```
deepansh96, line 73: Can you provide a doc link for this, for further reading?

src/main.js
```diff
@@ -55,6 +59,25 @@ const filterBeforeCreate = (toast, toasts) => {
 
 const app = createApp(App).use(store).use(router);
 
+if (
+  ["staging", "production"].includes(process.env.NODE_ENV) &&
```
deepansh96, line 63: In our frontend repo, where is the NODE_ENV set?
I checked in github secrets its not there

Will also need to pass these variables in the dockerfile and then in the deployment workflows.yml as well right?

src/main.js
rathorevaibhav: Done.

rathorevaibhav: I've added a comment within the code to refer more configs.

src/main.js
rathorevaibhav: I was under the impression that it gets sets automatically based on `npm run` or `npm build` command. Let me check.

docs/ERROR-MONITORING.md
```diff
@@ -0,0 +1,53 @@
...
+### Pre-requisites
+1. Create an account at Sentry.
+2. Set up your organization.
+3. Create a new VueJS project. Name it `plio-frontend`.
+
+
+### Enable error logging
+1. Set up env variable for `SENTRY_DSN`. You can get the value of DSN for your projects from `Project Settings > SDK Setup > Client Keys (DSN)`
+2. Make sure your `NODE_ENV` is either 'staging' or 'production'.
+3. Trigger an error by adding the following line in your code:
```
dalmia, line 14:
```suggestion
3. Test your set up by triggering an error - add the following line in your `urls.py` file:
```

docs/ERROR-MONITORING.md
```diff
@@ -0,0 +1,53 @@
...
+    ```js
+    throw "Deliberate Error!";
+    ```
+4. You should now see the error at your Sentry dashboard as well from `Issues` tab.
```
dalmia, line 18:
```suggestion
4. You should now see the error at your Sentry dashboard once you invoke the error trigger as well from `Issues` tab.
```

src/main.js
```diff
@@ -55,6 +59,26 @@ const filterBeforeCreate = (toast, toasts) => {
...
+  // Refer Sentry documentation for more configs: https://docs.sentry.io/platforms/javascript/configuration/
+  Sentry.init({
+    dsn: process.env.VUE_APP_SENTRY_DSN,
+    integrations: [new Integrations.BrowserTracing()],
+    // Set tracesSampleRate to 1.0 to capture 100%
+    // of transactions for performance monitoring.
+    tracesSampleRate: 1.0,
+    environment: process.env.NODE_ENV,
+  });
+  app.config.errorHandler = (err) => {
```
dalmia, line 77: any way to identify the user here?

src/main.js
```diff
@@ -55,6 +59,26 @@ const filterBeforeCreate = (toast, toasts) => {
...
+    Sentry.captureException(err);
```
dalmia, line 78: we also need to set this up for analytics too right?

src/main.js
rathorevaibhav: We can use the practices as mentioned below but have a look at my backend comment first about user privacy.
https://docs.sentry.io/platforms/javascript/data-management/sensitive-data/


Backend comment:
https://github.com/avantifellows/plio-backend/pull/202#discussion_r658587962

src/main.js
rathorevaibhav: I tried looking this for CubeJS but couldn't find a way yet. I have asked this on their forum as well and awaiting reply.

As it's mostly API calls, I believe we will still get to know from Frontend Sentry if an analytics API endpoint fails. However, to capture more info, we will need something at CubeJS end as well.
I'm keeping a track here: [Analytics issue](https://github.com/avantifellows/plio-analytics/issues/22)

src/main.js
deepansh96: https://cli.vuejs.org/guide/mode-and-env.html#modes
Vue CLI sets it for us! 
That's sorted then

src/main.js
dalmia: based on our conclusion on the backend, anything to update here?

src/main.js
dalmia: Cool. We can look at it later if we need to.

src/main.js
rathorevaibhav: Updated. I've set the user id now if present in vue store.

.env.example
```diff
@@ -36,3 +36,6 @@ VUE_APP_MIXPANEL_PROJECT_TOKEN= # get this from the Mixpanel project dashboard
 # chatwoot
 VUE_APP_CHATWOOT_URL= # URL of the deployed chatwoot app
 VUE_APP_CHATWOOT_TOKEN= # get this from your deployed chatwoot app's settings
+
+# Sentry Data Source Name
+VUE_APP_SENTRY_DSN=''
```
deepansh96, line 41:
```suggestion
VUE_APP_SENTRY_DSN=# Sentry Data Source Name
```
Just keeping it consistent with other values! Also if someone copies this file as their `.env` the DSN will be set to an empty string and the error reporting code will run

deepansh96 (changes requested):

PR #381, opened 2021-07-02
FIX: minor bugs

### LINKED ISSUE #379: When logging in with mobile number, chatwoot throws error when setting user
```js
// set the user for the chatwoot instance
window.$chatwoot.setUser(this.user.id, {
  email: this.user.email,
  name: this.user.first_name + this.user.last_name,
});
```
In the case when the user logs in with their mobile phone, the `this.user.email` is `null` and if `this.user.first_name` and `this.user.last_name` is null, then chatwoot throws an error as it NEEDS one of these keys to set the user.
Also, it doesn't accept empty string for those keys. 

Solution, if any of the keys are null, send a value like `"no_name"` or `"no_email"`

### LINKED ISSUE #380: Bug - method called on chatwoot variable when it's undefined
Screenshot 2021-07-02 at 9 38 38 PM

deepansh96:
Fixes #379 
Fixes #380

## Test Plan

- [ ] ~Test Responsiveness~
   - [ ] Laptop (1200px)
   - [ ] Tablet (760px)
   - [ ] Phone (320px)
- [ ] ~Cross-Browser Testing~
   - [ ] Chrome
   - [ ] Firefox
   - [ ] Safari
- [ ] ~Wrote tests~
- [x] Tested locally
- [ ] Tested on staging
- [ ] Tested on production

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -274,7 +274,7 @@ export default {
       // teardown chatwoot instance
 
       // reset the instance
-      window.$chatwoot.reset();
+      if (window.$chatwoot != undefined) window.$chatwoot.reset();
 
       // hide the bubble
       var chatwootBubble = document.querySelector(".woot-widget-bubble");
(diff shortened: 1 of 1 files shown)


src/App.vue
```diff
@@ -295,8 +295,8 @@ export default {
 
         // set the user for the chatwoot instance
         window.$chatwoot.setUser(this.user.id, {
-          email: this.user.email,
-          name: this.user.first_name + this.user.last_name,
+          email: this.user.email || "no_email",
```
dalmia, line 298: this will probably give error that `undefined has no attribute email`

deepansh96: No this piece of code runs only when `isAuthenticated` is true

PR #399, opened 2021-07-13
bugfix for changing workspace

sudheshna-donthineni:
The error occurred because this.currentPageNumber persists and the value is not reset. 
The hypothesis for the variable persisting is, it helps for searching a string on that particular page. 

**Solution:**
- Added a variable **_changedWorkspace_** in the watch property, **activeWorkspace**. 
- This is set true when fetchPlioIds() is called from the watch (i.e. when the user changes workspace)
- If so, the this.currentPageNumber is set to default.

### DIFF
--- a/src/pages/Home.vue
+++ b/src/pages/Home.vue
@@ -47,7 +47,10 @@
       </Paginator>
 
       <!-- no plios exist warning -->
-      <div v-if="!showTable" class="flex flex-col bg-white w-full m-auto mt-32 px-8">
+      <div
+        v-if="!showTable"
+        class="flex flex-col bg-white w-full m-auto mt-32 px-8"
+      >
         <inline-svg
           :src="noPliosIcon"
           class="w-50 h-50 opacity-50 place-self-center m-10"
@@ -90,7 +93,8 @@ export default {
   },
   watch: {
     async activeWorkspace() {
-      if (this.isUserApproved) await this.fetchPlioIds();
+      if (this.isUserApproved)
+        await this.fetchPlioIds({ changedWorkspace: true });
     },
     isUserApproved(value) {
       // fetch plios again if user approval status changes
@@ -172,10 +176,14 @@ export default {
       //if params contain a searchString or pageNumber, save it into a variable,
       //else save the variable as undefined
       var searchString =
-        params != undefined && "searchString" in params ? params.searchString : undefined;
+        params != undefined && "searchString" in params
+          ? params.searchString
+          : undefined;
 
       var pageNumber =
-        params != undefined && "pageNumber" in params ? params.pageNumber : undefined;
+        params != undefined && "pageNumber" in params
+          ? params.pageNumber
+          : undefined;
 
       // if the params contain a valid searchString, update the local searchString variable
       if (searchString != undefined && searchString != "")
@@ -183,6 +191,8 @@ export default {
 
       // if the params contain a valid pageNumber, update the local currentPageNumber variable
       if (pageNumber != undefined) this.currentPageNumber = pageNumber;
+
+      if (changedWorkspace) this.currentPageNumber = undefined;
 
       await PlioAPIService.getAllPlios(
         uuidOnly,
@@ -203,9 +213,11 @@ export default {
         })
         .then(async (plioIdList) => {
           // fetch the list of unique users for each plio
-          await PlioAPIService.getUniqueUsersCountList(plioIdList).then((response) => {
-            this.countUniqueUsersList = response;
-          });
+          await PlioAPIService.getUniqueUsersCountList(plioIdList).then(
+            (response) => {
+              this.countUniqueUsersList = response;
+            }
+          );
           return Promise.resolve(plioIdList);
         })
         .then((plioIdList) => this.prepareTableData(plioIdList)); // prepare the data for the table
@@ -242,7 +254,11 @@ export default {
         const plioId = plioIdList[plioIndex];
         var tableRow = {};
 
-        for (let colIndex = 0; colIndex < this.tableColumns.length; colIndex++) {
+        for (
+          let colIndex = 0;
+          colIndex < this.tableColumns.length;
+          colIndex++
+        ) {
           const column = this.tableColumns[colIndex];
           switch (column) {
             case "name":


src/pages/Home.vue
```diff
@@ -90,7 +93,8 @@ export default {
   },
   watch: {
     async activeWorkspace() {
-      if (this.isUserApproved) await this.fetchPlioIds();
+      if (this.isUserApproved)
+        await this.fetchPlioIds({ changedWorkspace: true });
```
dalmia, line 97: you can simply set `this.currentPageNumber = undefined;` here first and then call the function right?

src/pages/Home.vue
```diff
@@ -172,10 +176,14 @@ export default {
       //if params contain a searchString or pageNumber, save it into a variable,
       //else save the variable as undefined
       var searchString =
-        params != undefined && "searchString" in params ? params.searchString : undefined;
+        params != undefined && "searchString" in params
```
dalmia, line 179: Screenshot 2021-07-13 at 1 16 04 PM
can you set the line limit within your Vue linter to 90? OR else it will keep making these changes which are not relevant!

deepansh96: So the linter we use for vue files is called `vue 3 support - all in one`
You should install it from the extension store in vscode

src/pages/Home.vue
deepansh96: I think the best way to do this would be emit an event from the `WorkspaceSwitcher` component
Because we need to identify the case when the user has manually changed the workspace. There's a method in the `WorkspaceSwitcher` component called as `updateActiveWorkspace` which is invoked whenever the user switches the workspace manually.

We can emit an event there something like `"user-switched-workspace"` and listen to this event on `App.vue` and then unset the `currentPageNumber`

We should not do this inside the `watch` because there are other cases as well where the watch is called even when the user has not manually switched the workspace.

And I agree with aman that instead of passing a variable into `fetchPlioIds`, we should unset it right there and then

deepansh96: Also please mention the issue in the PR description like `Fixes #399`

sudheshna-donthineni: I could emit an event from Workspace Switcher and capture the event in App, but since the page number manipulation happens in Home, using the store was a better choice. I created a variable in the generic module, userSwitchedWorkspace which will be set to false initially. When the user manually changes the workspace, the mutation is done, setting it to true. After all the plios for that workspace is fetched, the variable is set back to false.

src/pages/Home.vue
```diff
@@ -263,6 +281,9 @@ export default {
       this.tableData = tableData;
       if (this.pending) this.stopLoading();
     },
+    resetPageNumber() {
+      this.currentPageNumber = undefined;
+    },
```
deepansh96, line 286: this method is not being used in the code

src/pages/Home.vue
```diff
@@ -242,7 +256,11 @@ export default {
         const plioId = plioIdList[plioIndex];
         var tableRow = {};
 
-        for (let colIndex = 0; colIndex < this.tableColumns.length; colIndex++) {
+        for (
+          let colIndex = 0;
+          colIndex < this.tableColumns.length;
+          colIndex++
+        ) {
```
deepansh96, line 263: did you install that linter?

src/pages/Home.vue
```diff
@@ -203,12 +214,15 @@ export default {
...
-          });
+          await PlioAPIService.getUniqueUsersCountList(plioIdList).then(
+            (response) => {
+              this.countUniqueUsersList = response;
+            }
+          );
           return Promise.resolve(plioIdList);
         })
         .then((plioIdList) => this.prepareTableData(plioIdList)); // prepare the data for the table
+      this.unsetUserSwitchedWorkspace();
```
deepansh96, line 225: I think this should come right after this check right?
```js
if (this.userSwitchedWorkspace) {
    this.currentPageNumber = undefined;
    this.unsetUserSwitchedWorkspace()
}
```

src/components/App/WorkspaceSwitcher.vue
```diff
@@ -43,7 +45,12 @@ export default {
         "Current Workspace": this.selectedWorkspace,
       });
       this.setActiveWorkspace(this.selectedWorkspace);
-      this.$router.push({ name: "Home", params: { org: this.activeWorkspace } });
+      this.$router.push({
+        name: "Home",
+        params: { org: this.activeWorkspace },
+      });
+
+      this.setUserSwitchedWorkspace();
```
deepansh96, line 53:
```suggestion
      this.setUserSwitchedWorkspace();
      
      this.$router.push({
        name: "Home",
        params: { org: this.activeWorkspace },
      });
```
Should ideally set the store variable before navigating to Home

src/components/App/WorkspaceSwitcher.vue
```diff
@@ -24,6 +24,7 @@ export default {
   name: "WorkspaceSwitcher",
   computed: {
     ...mapState("auth", ["activeWorkspace", "user"]),
+    ...mapState("generic", ["userSwitchedWorkspace"]),
```
deepansh96, line 27: not being used in this file

src/store/modules/generic.js
```diff
@@ -14,6 +15,12 @@ const actions = {
   unsetSharePlioDialog({ commit }) {
     commit("unsetSharePlioDialog");
   },
+  setUserSwitchedWorkspace({commit}){
+    commit("setUserSwitchedWorkspace")
+  }, 
+  unsetUserSwitchedWorkspace({commit}){
+    commit("unsetUserSwitchedWorkspace")
```
deepansh96, line 22: can you add one liner comments?

deepansh96 (changes requested):

src/components/App/WorkspaceSwitcher.vue
```diff
@@ -43,7 +44,11 @@ export default {
         "Current Workspace": this.selectedWorkspace,
       });
       this.setActiveWorkspace(this.selectedWorkspace);
-      this.$router.push({ name: "Home", params: { org: this.activeWorkspace } });
+      this.setUserSwitchedWorkspace(); //when the workspace is updated, setting this variable, so that it helps with the pagination of the table
```
deepansh96, line 47:
```suggestion
     // when the workspace is updated, set `userSwitchedWorkspace` store variable as `true`
      this.setUserSwitchedWorkspace();
```

deepansh96 (changes requested):

src/components/App/WorkspaceSwitcher.vue
```diff
@@ -35,6 +35,7 @@ export default {
   },
   methods: {
     ...mapActions("auth", ["setActiveWorkspace"]),
+    ...mapActions("generic", ["setUserSwitchedWorkspace"]),
```
deepansh96, line 38: not required if re-rendering is implemented right?

src/components/App/WorkspaceSwitcher.vue
```diff
@@ -43,11 +44,17 @@ export default {
         "Current Workspace": this.selectedWorkspace,
       });
       this.setActiveWorkspace(this.selectedWorkspace);
-      this.$router.push({ name: "Home", params: { org: this.activeWorkspace } });
+      this.$router.push({
+        name: "Home",
+        params: { org: this.activeWorkspace },
+      });
+      //event captured in App, helps in re-rendering Home component
```
deepansh96, line 51:
```suggestion
      // event captured in App.vue, helps in re-rendering the Home.vue component
```

src/pages/Home.vue
```diff
@@ -177,13 +177,21 @@ export default {
...
+      if (searchString != undefined && searchString != "") {
+        this.searchString = searchString;
+        this.currentPageNumber = undefined;
+      }
+
+      // if the user manually changed the workspace, reset the current page to default (or the beginning)
+      if (this.userSwitchedWorkspace) {
+        this.currentPageNumber = undefined;
+        this.unsetUserSwitchedWorkspace();
+      }
```
deepansh96, line 193: Not required anymore?

src/store/modules/generic.js
```diff
@@ -1,6 +1,7 @@
 const state = {
   isSharePlioDialogShown: false,
   plioLinkToShare: null,
+  userSwitchedWorkspace: false,
```
deepansh96, line 4: not required anymore?

src/store/modules/generic.js
```diff
@@ -14,6 +15,12 @@ const actions = {
   unsetSharePlioDialog({ commit }) {
     commit("unsetSharePlioDialog");
   },
+  setUserSwitchedWorkspace({ commit }) {
+    commit("setUserSwitchedWorkspace");
+  },
+  unsetUserSwitchedWorkspace({ commit }) {
+    commit("unsetUserSwitchedWorkspace");
+  },
```
deepansh96, line 23: not required anymore?

src/store/modules/generic.js
```diff
@@ -26,6 +33,13 @@ const mutations = {
   unsetSharePlioDialog(state) {
     state.isSharePlioDialogShown = false;
   },
+  //shows that the user manually changed the workspace using the Workspace Switcher
+  setUserSwitchedWorkspace(state) {
+    state.userSwitchedWorkspace = true;
+  },
+  unsetUserSwitchedWorkspace(state) {
+    state.userSwitchedWorkspace = false;
+  },
```
deepansh96, line 42: not required anymore?

deepansh96 (changes requested):

src/components/App/WorkspaceSwitcher.vue
sudheshna-donthineni: So, I removed it first, but then it started acting buggy, right? So, I wanted to see which change of mine was bringing that about, but couldn't test it. So waiting.

dalmia:
Closing in favor of #434

PR #428, opened 2021-09-01
Removed plio items query

rathorevaibhav:
Targets https://github.com/avantifellows/plio-backend/issues/232

## Summary

1. A separate items query no longer needed as we get items data within getPlio request
2. Corrections in mockResponse in Home.spec.js for delete plio functionality. Earlier the promises were not resolving and the test cases was getting finished.

## Test Plan

- [ ] ~Test Responsiveness~
   - [ ] Laptop (1200px)
   - [ ] Tablet (760px)
   - [ ] Phone (320px)
- [ ] ~Cross-Browser Testing~
   - [ ] Chrome
   - [ ] Firefox
   - [ ] Safari
- [ ] Local Language Support
- [x] Wrote tests
- [x] Tested locally
- [ ] Tested on staging
- [ ] Tested on production


src/services/API/Plio.js
```diff
@@ -27,16 +26,11 @@ export default {
       plioEndpoint += "/play";
     }
 
-    return Promise.all([
-      apiClient().get(plioEndpoint),
-      apiClient().get(itemsEndpoint, {
-        params: { plio: plioId },
-      }),
-    ]).then(([plio, items]) => {
+    return Promise.all([apiClient().get(plioEndpoint)]).then(([plio]) => {
```
dalmia, line 29: We don't need `.all()` now right? it was used because we were making multiple calls and wanted to wait for both of them to get resolved.

src/services/Testing/DummyData.js
```diff
@@ -1,3 +1,119 @@
+export const dummyItems = {
```
dalmia, line 1: has this just been moved up?

tests/unit/pages/Editor.spec.js
```diff
@@ -171,22 +171,17 @@ describe("Editor.vue", () => {
       },
```
dalmia, line 171: any reason you didn't delete the imported `dummyItems` here? 
Also, there is a `dummyItemsCopy` being created and used here - what about that?

rathorevaibhav:
@dalmia @deepansh96 this is ready for review again.

tests/unit/pages/Home.spec.js
```diff
@@ -369,13 +370,7 @@ describe("Home.vue", () => {
...
     await flushPromises();
 
-    mockAxios.mockResponse(
-      {
-        access_token: "",
-        expires_in: "",
-      },
-      mockAxios.queue()[0]
-    );
+    mockAxios.mockResponse(dummyPublishedPlio, mockAxios.queue()[0]);
```
dalmia, line 375: why was this change made? This is not related to items right?

tests/unit/pages/Editor.spec.js
```diff
@@ -188,9 +188,9 @@ describe("Editor.vue", () => {
 
     // use `wrapper.vm.__` to access the updated data variables inside the component
     expect(wrapper.vm.loadedPlioDetails.items).toStrictEqual(
-      dummyItemsCopy.data
```
dalmia, line 196: we should also remove the definition of `dummyItemsCopy` right?

tests/unit/pages/Editor.spec.js
```diff
@@ -171,31 +171,26 @@ describe("Editor.vue", () => {
...
     // using the fake response data
     mockAxios.mockResponse(plioResponse, mockAxios.queue()[0]);
-    mockAxios.mockResponse(itemResponse, mockAxios.queue()[1]);
 
     // wait until the DOM updates after promises resolve
     await flushPromises();
 
     // use `wrapper.vm.__` to access the updated data variables inside the component
     expect(wrapper.vm.loadedPlioDetails.items).toStrictEqual(
-      dummyItemsCopy.data
```
dalmia, line 196: There are many other places where `dummyItemsCopy` is being used - do search for it in the file and you'll find it - we should be replacing all of them right?

tests/unit/pages/Home.spec.js
rathorevaibhav: In home page here's what my understanding is for the flow when a plio is deleted:
1. A plio delete action is triggered - it makes a "DELETE" request.
2. A new query is made to get the list of all plios
3. For each plio, `getPlio` is requested

This line `373` is supposed to mock the getPlio response. The previous version of the `getPlio` had two promises inside `Promise.all()`, but because of the mockResponse here (with only one JSON), the promises didn't resolve. The test case finished without error.

In the new `getPlio`, as we only request for plio, we need a dummyPlio as the mock API response.

tests/unit/pages/Editor.spec.js
rathorevaibhav: I'm also not sure why we needed to have a clone of `dummyItems`. Maybe @deepansh96 have some insights here?

PR #441, opened 2021-09-03
Enable searching on pressing enter in the search box

### LINKED ISSUE #440: Run search after keyboard enter press

### LINKED ISSUE #442: Bugs with search and changing workspace
1. Keep 2 workspaces ready:
Personal - Have some plios
AF - Have no plios

In personal workspace, search something which gives no results -> we see the no plios found table

Now switch the workspace -> we see AF's workspace with no plios and the starting screen

Switch back to personal, it still shows the same no plios and the starting screen.

2. - Keep 2 workspaces ready
Personal - Have some plios
AF - Have no plios

In personal workspace, search something which gives SOME results -> we see the results filtered and shown to us

Now switch the workspace -> we see AF's workspace with no plios and the starting screen

Switch back to personal, it still shows the filtered results but in the search bar there is no string
- either we display that string there somehow
- or we dont' show them the filtered table and show them the default table

3. If you have 2 workspaces, both of which have some plios: if you search something in one workspace, then switch workspaces, the search string still remains there.

dalmia:
Fixes #440 
Fixes #442 

## Test Plan

- [ ] ~Test Responsiveness~
   - [ ] Laptop (1200px)
   - [ ] Tablet (760px)
   - [ ] Phone (320px)
- [x] Cross-Browser Testing
   - [ ] Chrome
   - [ ] Firefox
   - [ ] Safari
- [x] Wrote tests
- [x] Tested locally
- [ ] Tested on staging
- [ ] Tested on production

### DIFF
--- a/src/components/Collections/Table/Table.vue
+++ b/src/components/Collections/Table/Table.vue
@@ -304,7 +304,7 @@ export default {
          * event.key is the modern way of detecting keys
          * event.keyCode is deprecated (left here for for legacy browsers support)
          */
-        if (this.searchString != "") this.search();
+        if (this.searchString.trim() != "") this.search();
       }
     },
     resetSearchString() {
--- a/src/components/Collections/Table/Table.vue
+++ b/src/components/Collections/Table/Table.vue
@@ -19,9 +19,10 @@
         <!-- search bar input -->
         <input
           :class="searchInputBoxClass"
-          type="text"
           :placeholder="searchPlaceholder"
           v-model="searchString"
+          @keypress="searchIfEnter($event)"
+          type="text"
           autocomplete="off"
           data-test="searchBar"
         />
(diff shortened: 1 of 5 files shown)


src/components/Collections/Table/Table.vue
```diff
@@ -292,6 +293,20 @@ export default {
...
+       */
+      // check if the key pressed is the enter key
+      if (event.key === "Enter" || event.keyCode === 13) {
+        /**
+         * event.key is the modern way of detecting keys
+         * event.keyCode is deprecated (left here for for legacy browsers support)
+         */
+        if (this.searchString != "") this.search();
+      }
+    },
```
deepansh96, line 309: Are we sure that in legacy browsers, `event.key` will not throw an error like "property key of undefined not found"?

src/components/Collections/Table/Table.vue
```diff
@@ -292,6 +293,20 @@ export default {
...
+       * detect if enter has been pressed after entering
+       * a text to search
+       */
+      // check if the key pressed is the enter key
+      if (event.key === "Enter" || event.keyCode === 13) {
+        /**
+         * event.key is the modern way of detecting keys
+         * event.keyCode is deprecated (left here for for legacy browsers support)
+         */
+        if (this.searchString != "") this.search();
```
deepansh96, line 307: I think we should also check for leading or trailing spaces here and if they're present, trim those before proceeding with this check

deepansh96 (changes requested):

src/components/Collections/Table/Table.vue
dalmia: haan woh `event.key` sab mein hi hota hain from what I know!

src/components/Collections/Table/Table.vue
dalmia: makes sense

PR #489, opened 2021-10-26
FIX: better error handling for re-auth cases

### LINKED ISSUE #490: Error Handling - Re-authentication.
Because of some unhandled errors and exceptions, those errors were being encountered a lot of times by users and consecutively logged onto sentry.
Need to add error handling for the errors related to the re-authentication process

deepansh96:
Fixes #490 

## Test Plan

- [ ] ~Test Responsiveness~
   - [ ] Laptop (1200px)
   - [ ] Tablet (760px)
   - [ ] Phone (320px)
- [ ] ~Cross-Browser Testing`
   - [ ] Chrome
   - [ ] Firefox
- [ ] Wrote tests
- [x] Tested locally
- [x] Self-review
- [ ] Tested on staging
- [ ] Tested on production


src/services/Functional/User.js
```diff
@@ -13,7 +13,8 @@ export default {
       ) {
         store.dispatch("auth/unsetAccessToken");
         store.dispatch("auth/setReAuthenticationState", false);
-        return Promise.resolve(true);
+        store.dispatch("auth/unsetReAuthenticationPromise");
+        return Promise.reject("Auto Logout!");
```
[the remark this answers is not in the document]
dalmia: Got it. Thanks for explaining! 👍

src/services/Functional/User.js
```diff
@@ -13,7 +13,7 @@ export default {
       ) {
         store.dispatch("auth/unsetAccessToken");
         store.dispatch("auth/setReAuthenticationState", false);
-        return Promise.resolve(true);
+        return Promise.reject("Auto Logout!");
```
dalmia, line 16: Screenshot 2021-10-30 at 4 23 00 PM
I was trying to replicate this situation locally by intentionally setting isReAuthenticating to true and triggering the reAuthenticate method. I am getting this error - after the "Auto Logout" statement - as well!

deepansh96: - Haan ye second error will be resolved by that websocket waali PR that is open
- How did you try to replicate the situation? Because whenever I'm doing I'm never getting this
- Try once again, I've made some more fixes to handle errors better. Pehle the errors which were being originated at point A were lost somewhere in between. They were not still not being `caught` properly. (hence the red exceptions). I've tried to catch them
- Also, found a few more bugs

src/services/API/User.js
```diff
@@ -71,16 +71,19 @@ export default {
...
+        refreshTokenEndpoint,
+        {
+          grant_type: "refresh_token",
+          client_id: process.env.VUE_APP_BACKEND_API_CLIENT_ID,
+          client_secret: process.env.VUE_APP_BACKEND_API_CLIENT_SECRET,
+          refresh_token: store.state.auth.accessToken.refresh_token,
+        },
+        { baseURL: process.env.VUE_APP_BACKEND_AUTH_URL }
+      );
+    } else return Promise.reject("Refresh token not found in local storage");
```
dalmia, line 87: don't need else here right?

src/services/API/RootClient.js
```diff
@@ -99,7 +95,13 @@ client.interceptors.response.use(
             return client.request(error.config);
           }
         })
-        .catch((error) => console.log(error));
+        .catch((error) => {
+          console.log(error);
+          // auto logout the user
+          store.dispatch("auth/autoLogoutUser");
+          // return a never resolving/rejecting promise so no more API calls can occur
```
dalmia, line 102: question: meaning of this!

dalmia: also, why would this error happen in the first place?

src/services/API/RootClient.js
```diff
@@ -67,12 +67,8 @@ client.interceptors.response.use(
       error.config.url == refreshTokenEndpoint &&
       status === 400
     ) {
-      // unset the access token and log out the user
```
dalmia, line 70: also, why should this error occur? ideally, it should not occur right?

src/store/modules/auth.js
```diff
@@ -99,20 +102,22 @@ const actions = {
...
   async fetchAndUpdateUser({ dispatch, state }) {
-    await UserAPIService.getUserByAccessToken(state.accessToken.access_token)
-      .then(async (response) => {
-        await dispatch("setUser", response.data);
-      })
-      .catch((error) => console.log(error));
+    let response = await UserAPIService.getUserByAccessToken(
+      state.accessToken.access_token
+    );
+    if (response != undefined) dispatch("setUser", response.data);
```
dalmia, line 108: when would this be undefined?

src/services/Functional/User.js
```diff
@@ -11,21 +11,25 @@ export default {
...
-        store.dispatch("auth/unsetAccessToken");
-        store.dispatch("auth/setReAuthenticationState", false);
+        store.dispatch("auth/autoLogoutUser");
         return Promise.reject("Auto Logout!");
       } else {
         return store.state.auth.reAuthenticationPromise;
       }
     }
 
+    // if the reAuthentication call is made while the user has been logged out
```
dalmia, line 21: what is the case when this will happen?

src/services/Functional/User.js
```diff
@@ -11,21 +11,25 @@ export default {
         store.state.auth.reAuthenticationPromise == null ||
         UtilitiesService.isObjectEmpty(store.state.auth.reAuthenticationPromise)
       ) {
-        store.dispatch("auth/unsetAccessToken");
-        store.dispatch("auth/setReAuthenticationState", false);
-        return Promise.resolve(true);
+        store.dispatch("auth/autoLogoutUser");
+        return Promise.reject("Auto Logout!");
```
dalmia, line 15: I was still getting the red exception for "Auto Logout". Resolved it by using this instead. Read [this](https://stackoverflow.com/questions/44678505/uncaught-in-promise) to understand more. We were sending the rejected promise instead of the promise!

```python
return new Promise((reject) => {
          reject("Auto Logout!")
        });
```

To replicate, I made this change in `src/App.vue`:
Screenshot 2021-11-01 at 11 50 46 AM
Screenshot 2021-11-01 at 11 50 52 AM

PR #494, opened 2021-10-26
FIX: bug with users websocket connection when SSO user is updated

### LINKED ISSUE #493: Bug: Users websocket connection isn't updated when the logged in user is updated (using SSO)
Imagine this scenario -
- A user logs in to Plio using SAML SSO. The user id is `10`.
- The user id saved in localstorage is `10`
- A websocket connection is established to the channel for user `10`
- The user closes the browser
- A new user appears on the same browser with another SSO link
- Old user is logged out and new user is logged in.
- The user id saved in localstorage is updated to `11`
- But, here's the bug - The websocket connection still picks up the old saved user id. Even though the new user id is `11`, the websocket connection made is to the user `10`

deepansh96:
Fixes #493 

## Test Plan

- [ ] ~Test Responsiveness~
   - [ ] Laptop (1200px)
   - [ ] Tablet (760px)
   - [ ] Phone (320px)
- [ ] ~Cross-Browser Testing~
   - [ ] Chrome
   - [ ] Firefox
- [ ] ~Wrote tests~
- [x] Tested locally
- [x] Self-review
- [ ] Tested on staging
- [ ] Tested on production


src/store/plugins/usersWebSocketPlugin.js
```diff
@@ -51,11 +51,18 @@ export default function createUsersWebSocket() {
     const unsubscribe = store.subscribe((mutation) => {
       // subscribe to store mutations
 
-      // if the user has been newly set, make a new websocket connection and
-      // unsubscribe to the mutations of the previous websocket connection
-      if (mutation.type === "auth/setUser") connect(store, unsubscribe);
+      // if the user has been newly set, close any existing connection,
+      // make a new websocket connection and unsubscribe to the mutations of the previous websocket connection
+      if (mutation.type === "auth/setUser" && UsersWebsocket != undefined) {
```
dalmia, line 56: we are having to check for `UsersWebsocket != undefined` only because the code is proceeding after this line above:
```python
if (unsubscribeCallback != null) unsubscribeCallback();
```
We can simply return here after the unsubscribe callback is invoked and then, we won't need to place this check!

deepansh96: 1. `What is the situation in which subscribe is being triggered with UsersWebsocket undefined?`
- This happens when `auth/isAuthenticated` store variable is not set to true and the store observes a mutation. It doesn't matter what the mutation is. If `auth/isAuthenticated` is `false` and ANY mutation occurs, `UsersWebsocket` will be `undefined`. See the video below. I caught this bug when I was testing. In a test, the user was not authenticated and it showed that as something in the store mutated, it threw the `undefined` error.
https://user-images.githubusercontent.com/20169829/139599719-dde8f984-0e04-4dc8-9aee-528e0f07a059.mov

2. `Any reason that we are not simply returning the function right after this line?`
- Returning from the function will be wrong. Returning is not desired. We actually have to execute the following lines. When this line of code is run, that means that somewhere, a `connect()` call has been made. That call has been made with 2 arguments `store` and an `unsubscribe` callback. What this means is that, we want to configure the WS connection again, a fresh connection with fresh variables/objects. So it doesn't make sense to stop executing after the `unsubscribe` call. 
- 
- The `unsubscribe` call is made because if we don't do that, one more subscription will be registered to the store. If `connect` is called multiple times, multiple subscriptions would be made to the store. This is required because of the design of our method. 
- 
- The function to subscribe to the store is actually inside the `connect` method. So, to call the connect method again and again without keeping old (and not needed) subscriptions to the store, we need to unsubscribe to them.

dalmia: Screenshot 2021-11-01 at 11 16 21 AM
I don't think that it is because of ANY mutation. As the image shows, we have specifically set that the `connect` method should be invoked when the mutation is `auth/setUser`.

PR #509, opened 2021-11-03
FIX: better reauthentication logic

### LINKED ISSUE #508: Cycling reauthentication bug
**Describe the bug**
This is kind of an heisenbug. It is not exactly replicable, but if you try enough times, you should be able to see it. This bug happens during reAuthentication process. If someone's access token expires then a re authentication call is triggered.

Because of the way code was written, even when the call to retrieve a new valid access token was successful, the subsequent API calls were made using the old (invalid) access token. That, in turn, triggered more re authentication calls.

Every time an API call was made, it was made using the old access token whereas the process of getting a new access token was under way.
Also, the code was very complex so that needs to be simplified

**To Reproduce**
Open the player, expire the access token from tableplus and reload the player. You will need to do it multiple times. Keep the console open and maybe open a bunch of tabs on the browser so the browser is slowed down. You will see the recurring 401 error being logged in the console again and again.

This error was more evident when viewing a plio using SSO(maybe because of the extra loading time it takes)

**Screenshots**

https://user-images.githubusercontent.com/20169829/140056527-3cec4718-551d-46d2-9ef6-cd89d166a30c.mov

deepansh96:
Fixes #508 

## Solution
- Make sure you read the issue before going through the given solution
- Getting rid of the old store variable `isReauthenticating`
- Using a new store variable called `reAuthenticationState`. This store variable can have the below possible values
   -  `not-started` - This means that the re authentication process has not yet started or the process has finished and all the subsequent calls are now using the new refreshed access token
   - `in-process` - This means that currently a re authentication process is going on. 
   - `completed` - This means that a re authentication process was going on, and is now completed. All the pending requests that need a new access token can just use it and retry requesting again
   - `dropped` - This means that a re authentication process was started, but something went wrong and re authentication was not possible.
- Re using the store variable `reAuthenticationPromise`
- Using a new store variable called `reAuthenticationPromiseResolver` whose job is to resolve the `reAuthenticationPromise`.
- This is how the logic looks like
    - Whenever a request's response returns a 401 error, the code checks the value of `reAuthenticationState`
    - If `reAuthenticationState` is not started, then the method `UserFunctionalService.reAuthenticate(store)` is called
        - Inside the `reAuthenticate` method, we change the value of `reAuthenticationState` to `in-process`
        - We also create a promise, save it into the store. 
        - We also expose the resolver of the above promise and save it into the store as well. What this will allow us to do is we can wait for the promise somewhere else and resolve it from somewhere else.
        - Then we try to refresh the access token, and if we get a successful response, we store the access token into the store
        - we then set the `reAuthenticationState` to `completed`
        - we then resolve the `reAuthenticationPromise` - Why we do this, I'll explain later on
        - If there's some error while refreshing the access token, we throw an error (which will be caught at the callee - `RootClient.js`) and we set `reAuthenticationState` as `dropped`
    - If `reAuthenticationState` is `in-process`, then we need to wait for the process to complete. So we'll await the `reAuthenticationPromise` to fulfill. When the promise is resolved from inside the `UserFunctionalService.reAuthenticate(store)` method, then this await will resolve and we will move forward
    - If `reAuthenticationState` is `dropped`, then we throw an error. We will reset the `reAuthenticationState` to `not-started` so future re authentication processes can work correctly.

- ~[ ] Test Responsiveness~
   - [ ] Laptop (1200px)
   - [ ] Tablet (760px)
   - [ ] Phone (320px)
- ~[ ] Cross-Browser Testing~
   - [ ] Chrome
   - [ ] Firefox
- ~[ ] Local Language Support~
- [ ] Wrote tests
- [x] Tested locally
- [x] Self-review
- ~[ ] Check for bundle size [here](https://bundlephobia.com/) if adding a package~
- ~[ ] Images have `alt` attributes~
- ~[ ] Any `target="_blank"` links have `rel="noopener"`~
- ~[ ] Only SVGs are used as images. If PNGs are used, their size has been optimised.~
- [ ] Tested on staging
- ~[ ] Tested on an actual physical phone~
- [ ] Tested on production

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -146,9 +146,9 @@ export default {
   },
   async created() {
     // whenever the app is set up (on a fresh page load)
-    // reset the re authentication state. This is required because if in the
-    // previous run, if the user exited while the re authentication was in process,
-    // the re authentication state is set as `in-process` in the store and the next
+    // reset the reAuthenticationState. This is required because if in the
+    // previous run, if the user exited while the re-authentication was in process,
+    // the reAuthenticationState is set as `in-process` in the store and the next
     // time user reloads, the value will remain the same
     this.setReAuthenticationState("not-started");
 
--- a/src/App.vue
+++ b/src/App.vue
@@ -145,6 +145,13 @@ export default {
     };
   },
   async created() {
+    // whenever the app is set up (on a fresh page load)
+    // reset the re authentication state. This is required because if in the
+    // previous run, if the user exited while the re authentication was in process,
+    // the re authentication state is set as `in-process` in the store and the next
+    // time user reloads, the value will remain the same
+    this.setReAuthenticationState("not-started");
+
     // reset the value of pending while creating the component
     if (this.pending) this.stopLoading();
     // reset the value of whether background is disabled
@@ -247,6 +254,7 @@ export default {
       "unsetAccessToken",
       "fetchAndUpdateUser",
       "unsetActiveWorkspace",
+      "setReAuthenticationState",
     ]),
     ...mapActions("generic", [
       "unsetSharePlioDialog",
--- a/src/services/API/RootClient.js
+++ b/src/services/API/RootClient.js
@@ -53,7 +53,7 @@ client.interceptors.request.use(
 client.interceptors.response.use(
   (config) => {
     // reset reAuthenticationState to its original value whenever any request returns
-    // a non error response code. Make sure not to do this for refresh token call
+    // a non-error response code. Make sure not to do this for refresh token call
     if (config.config.url != refreshTokenEndpoint)
       store.dispatch("auth/setReAuthenticationState", "not-started");
     return config;
--- a/src/services/API/RootClient.js
+++ b/src/services/API/RootClient.js
@@ -133,7 +133,7 @@ client.interceptors.response.use(
       // the reason why we're not using the above try/catch logic as it as and why
       // we're wrapping it into an async function
       // https://itnext.io/error-handling-with-async-await-in-js-26c3f20bc06a
-      return run();
+      return handleReAuthentication();
     }
 
     ErrorHandling.handleAPIErrors(error);
--- a/src/services/API/RootClient.js
+++ b/src/services/API/RootClient.js
@@ -52,9 +52,13 @@ client.interceptors.request.use(
 // handle errors in responses for API requests
 client.interceptors.response.use(
   (config) => {
+    // reset reAuthenticationState to it's original value whenever any request returns
+    // a non error response code. Make sure not to do this for refresh token call
+    if (config.config.url != refreshTokenEndpoint)
+      store.dispatch("auth/setReAuthenticationState", "not-started");
     return config;
   },
-  (error) => {
+  async (error) => {
     // logging the error
     console.log(error);
 
@@ -77,34 +81,59 @@ client.interceptors.response.use(
 
     // Handle expired/deleted access token here
     if (status === 401) {
-      // reauthenticate the user, and until that is happenining, pause all other
-      // requests. Once the user is authenticated, release all the paused requests
-      // with the new token attached to the header
-      return UserFunctionalService.reAuthenticate(store)
-        .then(() => {
-          if (store.state.auth.accessToken != null) {
-            var newAccessToken = store.state.auth.accessToken.access_token;
-            // Add the new access token to the header of the request
-            error.config.headers["Authorization"] = `Bearer ${newAccessToken}`;
-
+      // reAuthenticate the user if this is the first reAuthentication call. While reAuthentication is going on
+      // if some other request ends up here, then wait for the reAuthentication process to finish. Once that is done,
+      // set the newly retrieved access token as an auth header and retry the request. Any errors in the reauthentication
+      // process are caught and the user is logged out.
+      let run = async () => {
+        try {
+          let reAuthenticationState = store.state.auth.reAuthenticationState;
+          switch (reAuthenticationState) {
+            case "dropped":
+              // re authentication process was dropped, throw an error.
+              throw new Error(
+                "Re-authentication failed, no more calls allowed"
+              );
+            case "in-process":
+              // re authentication process in progress. Wait for the promise to resolve,
+              // then proceed
+              await store.state.auth.reAuthenticationPromise;
+              break;
+            case "not-started":
+              // fresh re authentication process to begin now. Wait for the process to complete,
+              // then proceed
+              await UserFunctionalService.reAuthenticate(store);
+              break;
+          }
+          // Add the new access token recieved from the re authentication process
+          // to the header of the request and retry the request
+          let newAccessToken = store.state.auth.accessToken.access_token;
+          error.config.headers["Authorization"] = `Bearer ${newAccessToken}`;
+          if (error.config.url == userFromTokenEndpoint) {
             // If the call is fetching a user from an access token, we need to update the
             // request params with the new access token
-            if (error.config.url == userFromTokenEndpoint)
-              error.config.params["token"] = newAccessToken;
-
-            // try the request again
-            return client.request(error.config);
+            error.config.params["token"] = newAccessToken;
           }
-        })
-        .catch((error) => {
-          console.log(error);
-          // auto logout the user
-          store.dispatch("auth/autoLogoutUser");
-          // We want to cancel all the pending axios requests. The workaround for this is to
-          // return a never resolving/rejecting promise so no more API calls can occur
-          // https://github.com/axios/axios/issues/583#issuecomment-504317347
-          return new Promise(() => {});
-        });
+          return client.request(error.config); // retrying the request
+        } catch (err) {
+          // an error occured in the re authentication process
+          // reset the reAuthenticationState value, so any future reAuthentication processes
+          // can function normally
+          store.dispatch("auth/setReAuthenticationState", "not-started");
+          // if the user is still logged in, log them out
+          // if the user is not logged in, then just throw an error
+          if (store.getters["auth/isAuthenticated"]) {
+            store.dispatch("auth/autoLogoutUser");
+            throw new Error("AutoLogout! " + err.message);
+          }
+
+          throw new Error(err.message);
+        }
+      };
+      // the reason why we're not using the above try/catch logic as it as and why
+      // we're wrapping it into an async function
+      // https://itnext.io/error-handling-with-async-await-in-js-26c3f20bc06a
+      return run();
     }
 
     ErrorHandling.handleAPIErrors(error);
--- a/src/services/API/RootClient.js
+++ b/src/services/API/RootClient.js
@@ -81,7 +81,7 @@ client.interceptors.response.use(
 
     // Handle expired/deleted access token here
     if (status === 401) {
-      // reAuthenticate the user if this is the first reAuthentication call. While reAuthentication is going on
+      // re-authenticate the user if this is the first re-authentication call. While re-authentication is going on
       // if some other request ends up here, then wait for the reAuthentication process to finish. Once that is done,
       // set the newly retrieved access token as an auth header and retry the request. Any errors in the reauthentication
       // process are caught and the user is logged out.
--- a/src/services/API/RootClient.js
+++ b/src/services/API/RootClient.js
@@ -13,6 +13,12 @@ let headers = {
   "Content-Type": "application/json",
 };
 
+// whenever the RootClient is set up (on a fresh page load)
+// reset the re authentication state. This is required because if in the
+// previous run, if the user exited while the re authentication was in process,
+// the re authentication state is set as `in-process` in the store
+store.dispatch("auth/setReAuthenticationState", "not-started");
+
 // backend API client
 const client = axios.create({
   baseURL: process.env.VUE_APP_BACKEND,
--- a/src/services/API/RootClient.js
+++ b/src/services/API/RootClient.js
@@ -95,17 +95,15 @@ client.interceptors.response.use(
                 "Re-authentication failed, no more calls allowed"
               );
             case "in-process":
-              // re authentication process in progress. Wait for the promise to resolve,
-              // then proceed
+              // re-authentication process in progress. Wait for the promise to resolve,
               await store.state.auth.reAuthenticationPromise;
               break;
             case "not-started":
-              // fresh re authentication process to begin now. Wait for the process to complete,
-              // then proceed
+              // fresh re-authentication process to begin now. Wait for the process to complete,
               await UserFunctionalService.reAuthenticate(store);
               break;
           }
-          // Add the new access token recieved from the re authentication process
+          // Add the new access token recieved from the re-authentication process
           // to the header of the request and retry the request
           let newAccessToken = store.state.auth.accessToken.access_token;
           error.config.headers["Authorization"] = `Bearer ${newAccessToken}`;
@@ -116,8 +114,8 @@ client.interceptors.response.use(
           }
           return client.request(error.config); // retrying the request
         } catch (err) {
-          // an error occured in the re authentication process
-          // reset the reAuthenticationState value, so any future reAuthentication processes
+          // an error occured in the re-authentication process
+          // reset the reAuthenticationState value, so any future re-authentication processes
           // can function normally
           store.dispatch("auth/setReAuthenticationState", "not-started");
           // if the user is still logged in, log them out
--- a/src/services/API/RootClient.js
+++ b/src/services/API/RootClient.js
@@ -90,7 +90,7 @@ client.interceptors.response.use(
           let reAuthenticationState = store.state.auth.reAuthenticationState;
           switch (reAuthenticationState) {
             case "dropped":
-              // re authentication process was dropped, throw an error.
+              // re-authentication process was dropped, throw an error.
               throw new Error(
                 "Re-authentication failed, no more calls allowed"
               );
--- a/src/services/API/RootClient.js
+++ b/src/services/API/RootClient.js
@@ -54,7 +54,7 @@ client.interceptors.response.use(
   (config) => {
     // reset reAuthenticationState to its original value whenever any request returns
     // a non-error response code. Make sure not to do this for refresh token call
-    if (config.config.url != refreshTokenEndpoint)
+    if (config.config.url != refreshTokenEndpoint && store.state.auth.reAuthenticationState != "not-started")
       store.dispatch("auth/setReAuthenticationState", "not-started");
     return config;
   },
--- a/src/services/API/RootClient.js
+++ b/src/services/API/RootClient.js
@@ -82,7 +82,7 @@ client.interceptors.response.use(
     // Handle expired/deleted access token here
     if (status === 401) {
       // re-authenticate the user if this is the first re-authentication call. While re-authentication is going on
-      // if some other request ends up here, then wait for the reAuthentication process to finish. Once that is done,
+      // if some other request ends up here, then wait for the re-authentication process to finish. Once that is done,
       // set the newly retrieved access token as an auth header and retry the request. Any errors in the reauthentication
       // process are caught and the user is logged out.
       let run = async () => {
--- a/src/services/API/RootClient.js
+++ b/src/services/API/RootClient.js
@@ -85,7 +85,7 @@ client.interceptors.response.use(
       // if some other request ends up here, then wait for the re-authentication process to finish. Once that is done,
       // set the newly retrieved access token as an auth header and retry the request. Any errors in the re-authentication
       // process are caught and the user is logged out.
-      let run = async () => {
+      let handleReAuthentication = async () => {
         try {
           let reAuthenticationState = store.state.auth.reAuthenticationState;
           switch (reAuthenticationState) {
--- a/src/services/API/RootClient.js
+++ b/src/services/API/RootClient.js
@@ -13,12 +13,6 @@ let headers = {
   "Content-Type": "application/json",
 };
 
-// whenever the RootClient is set up (on a fresh page load)
-// reset the re authentication state. This is required because if in the
-// previous run, if the user exited while the re authentication was in process,
-// the re authentication state is set as `in-process` in the store
-store.dispatch("auth/setReAuthenticationState", "not-started");
-
 // backend API client
 const client = axios.create({
   baseURL: process.env.VUE_APP_BACKEND,
--- a/src/services/API/RootClient.js
+++ b/src/services/API/RootClient.js
@@ -52,7 +52,7 @@ client.interceptors.request.use(
 // handle errors in responses for API requests
 client.interceptors.response.use(
   (config) => {
-    // reset reAuthenticationState to it's original value whenever any request returns
+    // reset reAuthenticationState to its original value whenever any request returns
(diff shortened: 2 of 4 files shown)


src/App.vue
```diff
@@ -145,6 +145,13 @@ export default {
     };
   },
   async created() {
+    // whenever the app is set up (on a fresh page load)
+    // reset the re authentication state. This is required because if in the
+    // previous run, if the user exited while the re authentication was in process,
+    // the re authentication state is set as `in-process` in the store and the next
+    // time user reloads, the value will remain the same
+    this.setReAuthenticationState("not-started");
```
dalmia, line 153: if we are doing this, then why do we need `in-progress` at all?

src/services/Functional/User.js
```diff
@@ -1,51 +1,41 @@
...
 
 export default {
-  reAuthenticate(store) {
-    // return the stored promise from the re-authentication call if re-authentication is in process
-    if (store.state.auth.isReAuthenticating) {
-      // to handle the case when isReAuthenticating has been set to true
-      // but no promise exists in the store. Log out the user in that case
-      if (
-        store.state.auth.reAuthenticationPromise == null ||
-        UtilitiesService.isObjectEmpty(store.state.auth.reAuthenticationPromise)
```
dalmia, line 12: do we need this function within UtilitiesService anymore?

src/services/Functional/User.js
```diff
@@ -1,51 +1,41 @@
...
-    }
+  async reAuthenticate(store) {
+    // set re Authentication state as in progress
+    store.dispatch("auth/setReAuthenticationState", "in-process");
 
-    // if the reAuthentication call is made while the user has been logged out
-    // then reject the promise
-    if (!store.getters["auth/isAuthenticated"])
-      return Promise.reject("User is not set!");
+    // create a promise and store it's resolver in a separate variable
```
dalmia, line 8:
```suggestion
    // create a promise and store its resolver in a separate variable
```

src/services/Functional/User.js
```diff
@@ -1,51 +1,41 @@
...
+    let reAuthenticationPromiseResolver;
+    let reAuthenticationPromise = new Promise((resolve) => {
+      reAuthenticationPromiseResolver = resolve;
+    });
 
-    // set re-authentication as being in process
-    store.dispatch("auth/setReAuthenticationState", true);
+    // save the above created promise and it's resolver into the store
```
dalmia, line 14:
```suggestion
    // save the above created promise and its resolver into the store
```

src/services/Functional/User.js
```diff
@@ -1,51 +1,41 @@
...
-        UtilitiesService.isObjectEmpty(store.state.auth.reAuthenticationPromise)
-      ) {
-        store.dispatch("auth/autoLogoutUser");
-        return Promise.reject("Auto Logout!");
-      } else {
-        return store.state.auth.reAuthenticationPromise;
-      }
-    }
+  async reAuthenticate(store) {
+    // set re Authentication state as in progress
```
dalmia, line 5:
```suggestion
    // set reAuthentication state as in progress
```

src/services/Functional/User.js
```diff
@@ -1,51 +1,41 @@
...
-        // once the refreshAccessToken call resolves, set the new access token in store
-        // set isReAuthenticating to false
-        // and unset the reAuthenticationPromise that was saved in store
-        // and resolve the promise so other waiting calls can proceed
+    try {
+      // try refreshing the access token
+      let response = await UserAPIService.refreshAccessToken();
+      if (response != undefined && response.data != undefined) {
+        // set the recieved access token into the store
+        // and mark the re authentication process as completed.
```
dalmia, line 26:
```suggestion
        // and mark the re-authentication process as completed.
```

src/services/Functional/User.js
```diff
@@ -1,51 +1,41 @@
...
-    // store this promise into vuex and return from the function.
-    store.dispatch("auth/setReAuthenticationPromise", reAuthenticationPromise);
-    return reAuthenticationPromise;
+        store.dispatch("auth/setReAuthenticationState", "completed");
+        store.state.auth.reAuthenticationPromiseResolver();
+        return response.data;
+      }
+    } catch (e) {
+      // an error occured in the process of refreshing the access token.
+      // We'll mark the re authentication process as dropped and throw an error
```
dalmia, line 36:
```suggestion
      // We'll mark the re-authentication process as dropped and throw an error
```

src/services/Functional/User.js
```diff
@@ -1,51 +1,41 @@
...
+      store.dispatch("auth/setReAuthenticationState", "dropped");
+      throw new Error(e);
```
dalmia, line 38: we can use a PromiseRejector as well and use Promise.reject as well right? Would keep it consistent with the way we've been handling errors in general, no?

src/services/API/RootClient.js
```diff
@@ -52,9 +52,13 @@ client.interceptors.request.use(
 // handle errors in responses for API requests
 client.interceptors.response.use(
   (config) => {
+    // reset reAuthenticationState to it's original value whenever any request returns
```
dalmia, line 55:
```suggestion
    // reset reAuthenticationState to its original value whenever any request returns
```

src/services/API/RootClient.js
```diff
@@ -52,9 +52,13 @@ client.interceptors.request.use(
...
+    // a non error response code. Make sure not to do this for refresh token call
```
dalmia, line 56:
```suggestion
    // a non-error response code. Make sure not to do this for refresh token call
```

dalmia, line 56: why? `Make sure not to do this for refresh token call`

src/services/API/RootClient.js
```diff
@@ -77,34 +81,59 @@ client.interceptors.response.use(
...
-          if (store.state.auth.accessToken != null) {
-            var newAccessToken = store.state.auth.accessToken.access_token;
-            // Add the new access token to the header of the request
-            error.config.headers["Authorization"] = `Bearer ${newAccessToken}`;
-
+      // reAuthenticate the user if this is the first reAuthentication call. While reAuthentication is going on
+      // if some other request ends up here, then wait for the reAuthentication process to finish. Once that is done,
+      // set the newly retrieved access token as an auth header and retry the request. Any errors in the reauthentication
+      // process are caught and the user is logged out.
+      let run = async () => {
```
dalmia, line 88: better name for function, like, reAuthenticate or handleReAuthentication?

src/services/API/RootClient.js
```diff
@@ -77,34 +81,59 @@ client.interceptors.response.use(
...
+        try {
+          let reAuthenticationState = store.state.auth.reAuthenticationState;
+          switch (reAuthenticationState) {
+            case "dropped":
+              // re authentication process was dropped, throw an error.
```
dalmia, line 93: in some places, reAuthentication is written, in some places re authentication is written. Let's stick to one. I think re-authentication is fine.

src/services/API/RootClient.js
```diff
@@ -77,34 +81,59 @@ client.interceptors.response.use(
...
+              throw new Error(
+                "Re-authentication failed, no more calls allowed"
+              );
+            case "in-process":
+              // re authentication process in progress. Wait for the promise to resolve,
```
dalmia, line 98: re authentication inconsistency

src/services/API/RootClient.js
```diff
@@ -77,34 +81,59 @@ client.interceptors.response.use(
...
+              // then proceed
+              await store.state.auth.reAuthenticationPromise;
+              break;
+            case "not-started":
+              // fresh re authentication process to begin now. Wait for the process to complete,
```
dalmia, line 103: re authentication inconsistency

src/services/API/RootClient.js
```diff
@@ -77,34 +81,59 @@ client.interceptors.response.use(
...
+              // then proceed
+              await UserFunctionalService.reAuthenticate(store);
+              break;
+          }
+          // Add the new access token recieved from the re authentication process
```
dalmia, line 108: re authentication inconsistency

src/services/API/RootClient.js
```diff
@@ -77,34 +81,59 @@ client.interceptors.response.use(
...
-            if (error.config.url == userFromTokenEndpoint)
-              error.config.params["token"] = newAccessToken;
-
-            // try the request again
-            return client.request(error.config);
+            error.config.params["token"] = newAccessToken;
+          }
+          return client.request(error.config); // retrying the request
+        } catch (err) {
+          // an error occured in the re authentication process
```
dalmia, line 119: re authentication inconsistency

src/services/API/RootClient.js
```diff
@@ -77,34 +81,59 @@ client.interceptors.response.use(
...
+          // reset the reAuthenticationState value, so any future reAuthentication processes
```
dalmia, line 120: re authentication inconsistency

src/services/API/RootClient.js
```diff
@@ -77,34 +81,59 @@ client.interceptors.response.use(
...
-        .then(() => {
-          if (store.state.auth.accessToken != null) {
-            var newAccessToken = store.state.auth.accessToken.access_token;
-            // Add the new access token to the header of the request
-            error.config.headers["Authorization"] = `Bearer ${newAccessToken}`;
-
+      // reAuthenticate the user if this is the first reAuthentication call. While reAuthentication is going on
+      // if some other request ends up here, then wait for the reAuthentication process to finish. Once that is done,
+      // set the newly retrieved access token as an auth header and retry the request. Any errors in the reauthentication
+      // process are caught and the user is logged out.
```
dalmia, line 87: re authentication inconsistency

src/services/API/RootClient.js
```diff
@@ -77,34 +81,59 @@ client.interceptors.response.use(
...
+          let reAuthenticationState = store.state.auth.reAuthenticationState;
+          switch (reAuthenticationState) {
+            case "dropped":
+              // re authentication process was dropped, throw an error.
+              throw new Error(
+                "Re-authentication failed, no more calls allowed"
+              );
+            case "in-process":
+              // re authentication process in progress. Wait for the promise to resolve,
+              // then proceed
```
dalmia, line 99: this seems redundant

src/services/API/RootClient.js
```diff
@@ -77,34 +81,59 @@ client.interceptors.response.use(
...
+              await store.state.auth.reAuthenticationPromise;
+              break;
+            case "not-started":
+              // fresh re authentication process to begin now. Wait for the process to complete,
+              // then proceed
```
dalmia, line 104: this seems redundant

src/services/Functional/User.js
```diff
@@ -1,51 +1,41 @@
...
-      })
-      .catch((error) => {
-        return Promise.reject(error);
-      });
-
-    // Until the promise stored in "reAuthenticationPromise" is resolved,
-    // store this promise into vuex and return from the function.
-    store.dispatch("auth/setReAuthenticationPromise", reAuthenticationPromise);
-    return reAuthenticationPromise;
+        store.dispatch("auth/setReAuthenticationState", "completed");
```
dalmia, line 30: i don't see this state being used anywhere but maybe you wanted to log a state which was in-between in-progress, nor not-started? still, just think about its usage once.

src/services/API/RootClient.js
```diff
@@ -77,34 +81,59 @@ client.interceptors.response.use(
...
+        } catch (err) {
+          // an error occured in the re authentication process
+          // reset the reAuthenticationState value, so any future reAuthentication processes
+          // can function normally
+          store.dispatch("auth/setReAuthenticationState", "not-started");
+          // if the user is still logged in, log them out
+          // if the user is not logged in, then just throw an error
+          if (store.getters["auth/isAuthenticated"]) {
+            store.dispatch("auth/autoLogoutUser");
+            throw new Error("AutoLogout! " + err.message);
```
dalmia, line 127: what about?

```
// We want to cancel all the pending axios requests. The workaround for this is to
// return a never resolving/rejecting promise so no more API calls can occur
// https://github.com/axios/axios/issues/583#issuecomment-504317347
return new Promise(() => {});
```

dalmia, line 127: are you using throw now instead of Promise.reject because we are not returning any Promise anymore?

src/services/API/RootClient.js
deepansh96: Will explain on call

src/services/API/RootClient.js
```diff
@@ -77,34 +81,59 @@ client.interceptors.response.use(
...
-      // requests. Once the user is authenticated, release all the paused requests
-      // with the new token attached to the header
-      return UserFunctionalService.reAuthenticate(store)
-        .then(() => {
-          if (store.state.auth.accessToken != null) {
-            var newAccessToken = store.state.auth.accessToken.access_token;
-            // Add the new access token to the header of the request
-            error.config.headers["Authorization"] = `Bearer ${newAccessToken}`;
-
+      // reAuthenticate the user if this is the first reAuthentication call. While reAuthentication is going on
```
deepansh96, line 84:
```suggestion
      // re-authenticate the user if this is the first re-authentication call. While re-authentication is going on
```

src/services/API/RootClient.js
```diff
@@ -77,34 +81,59 @@ client.interceptors.response.use(
...
+      // if some other request ends up here, then wait for the reAuthentication process to finish. Once that is done,
```
deepansh96, line 85:
```suggestion
      // if some other request ends up here, then wait for the re-authentication process to finish. Once that is done,
```

src/services/API/RootClient.js
```diff
@@ -77,34 +81,59 @@ client.interceptors.response.use(
...
+      // set the newly retrieved access token as an auth header and retry the request. Any errors in the reauthentication
```
deepansh96, line 86:
```suggestion
      // set the newly retrieved access token as an auth header and retry the request. Any errors in the re-authentication
```

src/services/API/RootClient.js
deepansh96:
```suggestion
      let handleReAuthentication = async () => {
```

src/services/API/RootClient.js
```diff
@@ -77,34 +81,59 @@ client.interceptors.response.use(
...
-          return new Promise(() => {});
-        });
+
+          throw new Error(err.message);
+        }
+      };
+      // the reason why we're not using the above try/catch logic as it as and why
+      // we're wrapping it into an async function
+      // https://itnext.io/error-handling-with-async-await-in-js-26c3f20bc06a
+      return run();
```
deepansh96, line 136:
```suggestion
      return handleReAuthentication();
```

src/App.vue
```diff
@@ -145,6 +145,13 @@ export default {
     };
   },
   async created() {
+    // whenever the app is set up (on a fresh page load)
+    // reset the re authentication state. This is required because if in the
+    // previous run, if the user exited while the re authentication was in process,
+    // the re authentication state is set as `in-process` in the store and the next
+    // time user reloads, the value will remain the same
```
deepansh96, line 152:
```suggestion
    // whenever the app is set up (on a fresh page load)
    // reset the reAuthenticationState. This is required because if in the
    // previous run, if the user exited while the re-authentication was in process,
    // the reAuthenticationState is set as `in-process` in the store and the next
    // time user reloads, the value will remain the same
```

src/services/API/RootClient.js
deepansh96:
```suggestion
              // re-authentication process was dropped, throw an error.
```

src/services/API/RootClient.js
deepansh96:
```suggestion
              // re-authentication process in progress. Wait for the promise to resolve,
```

src/services/API/RootClient.js
deepansh96:
```suggestion
```

src/services/API/RootClient.js
deepansh96:
```suggestion
              // fresh re-authentication process to begin now. Wait for the process to complete,
```

src/services/API/RootClient.js
deepansh96:
```suggestion
```

src/services/API/RootClient.js
deepansh96:
```suggestion
          // Add the new access token recieved from the re-authentication process
```

src/services/API/RootClient.js
deepansh96:
```suggestion
          // an error occured in the re-authentication process
```

src/services/API/RootClient.js
deepansh96:
```suggestion
          // reset the reAuthenticationState value, so any future re-authentication processes
```

src/services/API/RootClient.js
deepansh96: We can do that as well. It doesn't really matter in this case because the user would be auto logged out.
Throwing an error will just allow it to be caught at the location where the API call is made.
For example, if the API call was made in Plio.vue
```js
await PlioAPIService.getPlio(...)
    .then(() => { ... })
    .catch( (e) => console.log(e) )
```
This will be logged here and that's it.

As the user will be logged out, no more API call would be scheduled anyway.

deepansh96: `async` functions always return a promise. Always.
Refer to the image below

image


So if it was not an async function, I would've written 
```js
return Promise.reject("Autologout")
```

But because the logic is wrapped in an async function, I can just throw an error and it will be wrapped in a promise and that promise will be rejected.

And why are we returning a promise anyway? Turns out we HAVE to return a promise from the interceptor. (you can try returning a non promise and it'll throw an error). We HAVE to return a promise from the interceptor as that is what's being expected by our originator invocations. (the originator invocations are the ones where we make a call like PlioAPISerivce.getPlio())

src/services/Functional/User.js
deepansh96: Not in this file, but this paritcular function is very helpful. It's just two lines and I wouldn't wanna search for this logic again from StackOverflow and write it. So i would suggest let it be

src/services/Functional/User.js
deepansh96: Yeah correct. It's not being used directly but if I HAD to write it, it would be kept here
```js
let reAuthenticationState = store.state.auth.reAuthenticationState;
          switch (reAuthenticationState) {
            case "dropped":
              // re-authentication process was dropped, throw an error.
              throw new Error(
                "Re-authentication failed, no more calls allowed"
              );
            case "in-process":
              // re authentication process in progress. Wait for the promise to resolve,
              // then proceed
              await store.state.auth.reAuthenticationPromise;
              break;
            case "not-started":
              // fresh re authentication process to begin now. Wait for the process to complete,
              // then proceed
              await UserFunctionalService.reAuthenticate(store);
              break;
            case "completed":
             break;
          }
```

src/services/Functional/User.js
deepansh96: Again, as this is an async function, it will always return a promise (refer to the image attached above in one comment).

If it was not an async function, we would've written Promise.reject(...).
It's not about being consistent actually here. This is the correct way.

src/store/modules/auth.js
```diff
@@ -89,14 +90,14 @@ const actions = {
...
     await commit("saveConfig", config);
   },
-  setReAuthenticationState({ commit }, isReAuthenticating) {
-    commit("setReAuthenticationState", isReAuthenticating);
+  async setReAuthenticationState({ commit }, state) {
+    await commit("setReAuthenticationState", state);
   },
-  setReAuthenticationPromise({ commit }, reAuthenticationPromise) {
-    commit("setReAuthenticationPromise", reAuthenticationPromise);
+  async setReAuthenticationPromise({ commit }, promise) {
```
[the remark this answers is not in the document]
deepansh96: I'm not sure, I think we can remove this. I just noticed the `async getAnalyticsAccessToken` written there. But I think we can remove it.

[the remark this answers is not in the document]
deepansh96: Read the async/await part here https://next.vuex.vuejs.org/guide/actions.html#composing-actions

I've removed useless async/await keywords in the store's actions

src/store/modules/auth.js
```diff
@@ -67,37 +67,37 @@ const getters = {
 
 const actions = {
   async setAccessToken({ commit, dispatch }, accessToken) {
-    await commit("setAccessToken", accessToken);
+    commit("setAccessToken", accessToken);
```
dalmia, line 70: why are we changing this? `fetchAndUpdateUser` is using `state.access_token` right? so, if is called before the access token is set, it will result in an error na?

src/store/modules/auth.js
```diff
@@ -67,37 +67,37 @@ const getters = {
...
   },
-  async setReAuthenticationPromise({ commit }, promise) {
-    await commit("setReAuthenticationPromise", promise);
+  setReAuthenticationPromise({ commit }, promise) {
+    commit("setReAuthenticationPromise", promise);
   },
-  async setReAuthenticationPromiseResolver({ commit }, resolver) {
-    await commit("setReAuthenticationPromiseResolver", resolver);
+  setReAuthenticationPromiseResolver({ commit }, resolver) {
+    commit("setReAuthenticationPromiseResolver", resolver);
```
dalmia: have all these functions being tested in the sense that there is no error after removing the async await - as in logically, as things following each of these calls waiting on them to be completed first? if yes, we should not remove the async/awaits right?

src/services/Functional/User.js
```diff
@@ -2,16 +2,16 @@ import UserAPIService from "@/services/API/User.js";
 
 export default {
   async reAuthenticate(store) {
-    // set re Authentication state as in progress
+    // set reAuthentication state as in progress
```
dalmia, line 5:
```suggestion
    // set re-authentication state as in progress
```

src/services/Functional/User.js
dalmia: We can always find it from our Github history if we need to, no?

src/services/Functional/User.js
dalmia: Didn't understand what you mean to say here. Will call!

src/services/API/RootClient.js
dalmia: But you cannot return a Promise inside a non-async function, right? So, what does this line mean?
Screenshot 2021-11-08 at 9 37 40 AM

Either we should use the `Promise.reject/resolve` thing everything or `return value / throw Error` thing everywhere.  Following both the styles will be confusing for anyone reading the code.

src/services/API/RootClient.js
```diff
@@ -52,9 +52,13 @@ client.interceptors.request.use(
 // handle errors in responses for API requests
 client.interceptors.response.use(
   (config) => {
+    // reset reAuthenticationState to it's original value whenever any request returns
+    // a non error response code. Make sure not to do this for refresh token call
+    if (config.config.url != refreshTokenEndpoint)
```
dalmia, line 57: should additionally place a check that the reAuthenticationState is not started

deepansh96:
```suggestion
    if (config.config.url != refreshTokenEndpoint && store.state.auth.reAuthenticationState != "not-started")
```

PR #523, opened 2021-11-11
Adds a side menu for the app

### LINKED ISSUE #520: Add a sidebar on the home page

dalmia:
Fixes #520 

## Summary
- removed the logo
- added a hamburger icon for menu
- added a menu bar for the app
- the menu stays open and static for the home page
- the menu opens as an overlay for non-home pages
- have added click-away for non-home pages as we have decided to keep the sidebar always on for the home page
- for screen-sizes < 500, the menu covers the whole screen when opened
- the menu remains closed by default if we directly land on a non-home page
- the home page used to shift because the search icon took some time to load - prevented that shifting.
- changed "Number of viewers" to "views" as they convey the same thing and the latter takes much lesser space (important to maintain good visibility along with the sidebar for screen-sizes between 500-900).
- some places had `transition ease` written but `ease` is not a valid tailwind class - so, have removed `ease` from the codebase.
-  using a single value for window's inner height and width for the entire app - kept within store.
- changed paginator to convert to previous/next mode starting from screen size 1024px to accommodate the menu bar being open for screen sizes > 500px.
- made the editor keep the split-screen layout until 500px (instead of only until 768px)

## Test Plan

- [x] Test Responsiveness
   - [ ] Laptop (1200px)
   - [ ] Tablet (760px)
   - [ ] Phone (320px)
- [x] Cross-Browser Testing
   - [ ] Chrome
   - [ ] Firefox
- [x] Local Language Support
- [ ] Wrote tests
- [x] Tested locally
- [ ] Self-review
- [x] Only SVGs are used as images. If PNGs are used, their size has been optimised.
- [x] Any SVGs without text have their `aria-label` attributes set
- [ ] Tested on staging
- [ ] Tested on production


src/App.vue
```diff
@@ -6,73 +6,110 @@
...
+          <icon-button
+            class="place-self-start"
+            :iconConfig="logoutButtonIconConfig"
+            :titleConfig="logoutButtonTextConfig"
+            :buttonClass="menuButtonsClass"
+            @click="logoutButtonClicked"
+            :isDisabled="pending"
+          ></icon-button>
         </div>
+        <router-view :class="routerViewClass" :key="$route.fullPath" />
```
deepansh96, line 111: Why is the key required here?

src/App.vue
```diff
@@ -132,17 +169,24 @@ import { useToast } from "vue-toastification";
...
   },
   data() {
     return {
       showLanguagePickerDialog: false, // whether to show a language picker dialog box
       toast: useToast(), // use the toast component
       userClickedLogout: false, // if the user has clicked the logout button
+      // class for the create button
+      createButtonClass:
+        "bg-primary hover:bg-primary-hover rounded-lg w-full ring-primary p-2 sm:py-4",
+      isMenuButtonPressed: false,
```
deepansh96, line 186:
```suggestion
      isMenuButtonActive: false,
```
Pressed feels like ki it's continuously pressed, like I've continuously kept my mouse clicked at it. Active seems more intuitive to me

src/App.vue
```diff
@@ -268,26 +316,28 @@ export default {
...
-        });
-        // resets the distinct ID so that multiple users can use the same device
-        this.$mixpanel.reset();
-        this.$mixpanel.track("Logout");
-        // added here so that if someone clicks on logout while
-        // some activity is pending
-        this.stopLoading();
-      });
+    redirectToHome() {
+      if (window.innerWidth <= 500) this.resetMenuPressedState();
```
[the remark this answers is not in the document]
deepansh96: Or
```suggestion
      if (window.innerWidth <= 500) this.resetMenuActiveState();
```

src/App.vue
```diff
@@ -312,6 +362,28 @@ export default {
           }
         })
         .catch(() => this.toast.error(this.$t("error.create_plio")));
+      this.resetMenuPressedState();
+    },
+    logoutButtonClicked() {
+      // set whether the logout action as triggered by the user or not
```
deepansh96, line 368:
```suggestion
      // set whether the logout action was triggered by the user or not
```

src/App.vue
```diff
@@ -361,37 +433,156 @@ export default {
...
+      // config for the plio docs button
+      return {
+        value: this.$t("nav.docs"),
+        class: this.menuButtonsTextClass,
+      };
     },
     showWorkspaceSwitcher() {
       // whether to show workspace switcher
-      return this.isAuthenticated && this.onHomePage && this.user.organizations.length;
+      return this.user.organizations.length > 0;
```
deepansh96, line 572: If `isAuthenticated` is false, then `this.user` will be null and this will throw an error no? saying that cannot find property `organizations` of `null`?

src/assets/images/docs.svg
```diff
@@ -0,0 +1 @@
+<svg height='100px' width='100px'  fill="#000000" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" version="1.1" x="0px" y="0px" viewBox="0 0 100 100" enable-background="new 0 0 100 100" xml:space="preserve"><g><polygon points="25.205,85.289 25.205,12.954 21.447,12.954 21.447,89.047 76.898,89.047 76.898,85.289  "></polygon><polygon points="18.543,91.476 18.543,19.14 14.785,19.14 14.785,95.234 70.236,95.234 70.236,91.476  "></polygon></g><polygon points="65.061,5.04 85.229,25.208 65.061,25.208 "></polygon><g><path d="M63.209,27.06V5.246H28.635v77.379h56.388V27.06H63.209z M76.671,67.823H36.986v-2.916h39.685V67.823z M76.671,56.371   H36.986v-2.916h39.685V56.371z M76.671,44.919H36.986v-2.916h39.685V44.919z"></path></g></svg>
```
deepansh96, line 1: Forgot to remove height, width and fill

src/assets/images/logout.svg
```diff
@@ -0,0 +1 @@
+<svg height='100px' width='100px'  fill="#000000" xmlns="http://www.w3.org/2000/svg" data-name="Layer 1" viewBox="0 0 80 80" x="0px" y="0px"><path d="M71,23.31V56.69a8.44,8.44,0,0,1-8.43,8.43H50a3.93,3.93,0,1,1,0-7.85H62.57a.58.58,0,0,0,.58-.58V23.31a.58.58,0,0,0-.58-.58H50a3.93,3.93,0,1,1,0-7.85H62.57A8.44,8.44,0,0,1,71,23.31Z"></path><path d="M51.38,40.49a3.93,3.93,0,0,1-3.92,3.93H22.39l8,7.95a3.91,3.91,0,0,1,0,5.55,3.91,3.91,0,0,1-5.54,0L10.14,43.27a4,4,0,0,1,0-5.56L24.8,23.07a3.93,3.93,0,0,1,6.7,2.77,3.9,3.9,0,0,1-1.16,2.78l-7.95,8H47.46A3.93,3.93,0,0,1,51.38,40.49Z"></path></svg>
```
deepansh96, line 1: Forgot to remove height, width and fill

src/assets/images/menu.svg
```diff
@@ -0,0 +1 @@
+<svg height='100px' width='100px'  fill="#000000" xmlns="http://www.w3.org/2000/svg" data-name="Layer 1" viewBox="0 0 100 100" x="0px" y="0px"><title>Essential Icons</title><path d="M80,28H20a3,3,0,0,1,0-6H80a3,3,0,0,1,0,6Z"></path><path d="M80,53H20a3,3,0,0,1,0-6H80a3,3,0,0,1,0,6Z"></path><path d="M80,78H20a3,3,0,0,1,0-6H80a3,3,0,0,1,0,6Z"></path></svg>
```
deepansh96, line 1: Forgot to remove height, width and fill

deepansh96 (changes requested):

src/App.vue
```diff
@@ -6,73 +6,110 @@
...
+          class="grid justify-items-end bp-500:col-span-2"
+          :class="{ 'col-span-2': !onHomePage }"
         >
+          <!-- locale switcher -->
+          <LocaleSwitcher id="locale" class="flex justify-center"></LocaleSwitcher>
+        </div>
+      </div>
+      <!-- main container -->
+      <div :class="gridContainerClass">
+        <div :class="menuContainerClass" v-if="isMenuShown" class="menu">
```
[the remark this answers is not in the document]
dalmia: I was trying to add that transition by myself and this was added as a part of that experimentation - removing it!

src/App.vue
```diff
@@ -361,37 +433,156 @@ export default {
...
+          "absolute z-10 bg-white w-full bp-500:w-auto": this.isMenuShownOverlay,
+        },
+        `p-2 sm:p-4 border-r-2 flex flex-col h-screen col-span-3 bp-500:col-span-1 h-screen`,
+      ];
+    },
+    gridContainerClass() {
       return {
-        hidden: this.isNavBarHidden,
+        "grid grid-cols-3 lg:grid-cols-4 xl:grid-cols-5": this.isMenuShownSideways,
+        flex: this.isMenuShownOverlay,
```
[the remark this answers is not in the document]
dalmia: when the class is just one word, the formatter automatically removes the `""`.

src/App.vue
dalmia: So, there was a bug before this: if you are in the editor and you click the create button from the menu, the URL changes but since the component doesn't change, it wouldn't reload the page.

Adding this lets Vue know that it should reload even if the component is the same.

src/App.vue
```diff
@@ -6,73 +6,110 @@
...
+          :iconConfig="menuButtonIconConfig"
+          :buttonClass="menuButtonClass"
+          class="rounded-md border shadow-lg"
+          @click="toggleMenuButton"
+          :isDisabled="pending"
+        ></icon-button>
 
         <!-- create plio button -->
+        <icon-button
+          v-if="onHomePage"
```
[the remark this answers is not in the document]
dalmia: `on nav bar` seems redundant to me as it is inside the nav bar div? Adding the rest though!

src/App.vue
```diff
@@ -6,73 +6,110 @@
...
-            class="flex justify-center"
-            :isDisabled="pending"
-          ></WorkspaceSwitcher>
-        </div>
+        <!-- menu icon -->
+        <icon-button
+          v-if="isAuthenticated"
+          :iconConfig="menuButtonIconConfig"
+          :buttonClass="menuButtonClass"
+          class="rounded-md border shadow-lg"
```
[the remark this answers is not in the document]
dalmia: `rounded-lg` comes from the default classes we have for the button. We want it to be overridden by whatever we provide later. But haan can move the classes to one place only.

src/App.vue
dalmia: if `isAuthenticated` is false, the whole menu won't be rendered in the first place - this will only be evaluated when the menu is rendered - the `isAuthenticated` check there takes care of it and doesn't have to be repeated here.

src/App.vue
```diff
@@ -212,6 +215,11 @@ export default {
     // add event listeners to track if network connection went up or down
     window.Offline.on("down", this.showInternetLostToast);
     window.Offline.on("up", this.showInternetRestoredToast);
+
+    // if user lands on home page, no transition should be applied
+    // else apply a transition with the name "slide-fade"
+    if (this.onHomePage) this.menuSlideTransition = "";
+    else this.menuSlideTransition = "slide-fade";
```
dalmia, line 222: just setting this in the watcher will suffice.

dalmia: only this case needs to be written here, the case for this.onHomePage being true suffices from being in the watcher itself

src/App.vue
```diff
@@ -470,21 +480,24 @@ export default {
...
           "opacity-50 pointer-events-none w-full": this.isMenuShownOverlay,
         },
       ];
     },
     menuContainerClass() {
       return [
         {
-          "absolute z-10 bg-white w-full bp-500:w-auto": this.isMenuShownOverlay,
+          "absolute z-10 bg-white w-full bp-500:w-1/4":
+            this.isMenuShownOverlay || (!this.isMenuShown && !this.onHomePage),
```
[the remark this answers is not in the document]
deepansh96: This change and the change below both go hand in hand. These both changes are needed as without these, these two issues were happening

https://user-images.githubusercontent.com/20169829/141514892-90ca2002-db69-443f-a610-12ddc979b0e9.mov

Assuming I hadn't made the changes - 
In the above recording, you can see that if I click the menu button on home page (when menu button was still there, now it's removed), it takes some time to go away due to the transition. While it's going away, our variables for `isMenuShown` change.  The `gridContainerClass` would lose it's `grid` classes. The `menuContainerClass` would occupy the whole screen (just for the split second of time while the transition is taking place). So we need to retain the `gridContainerClass` ka grid classes and also restrict the menu bar to retain it's size when transitioning.

One more issue happens when switching from home to some other page -- and the reason for it is the same as above



https://user-images.githubusercontent.com/20169829/141518797-62b1b5a9-7ecd-497f-91bb-a60de948e937.mov

src/App.vue
```diff
@@ -470,21 +480,24 @@ export default {
...
         },
-        `p-2 sm:p-4 border-r-2 flex flex-col h-screen col-span-3 bp-500:col-span-1`,
+        `p-2 sm:p-4 border-r-2 flex flex-col h-screen-adjusted col-span-3 bp-500:col-span-1`,
       ];
     },
     gridContainerClass() {
       return {
-        "grid grid-cols-3 lg:grid-cols-4 xl:grid-cols-5": this.isMenuShownInline,
+        "grid grid-cols-3 lg:grid-cols-4 xl:grid-cols-5":
+          this.isMenuShownInline || !this.isMenuShown,
```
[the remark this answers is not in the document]
deepansh96: Described above

deepansh96:
Can you now add click-away?
Because now the menu will be anyway visible and static on the home page, having a click away on the menu would be helpful in cases of non-home pages - where the menu is shown as an overlay and whenever we click outside we can see the menu close up

src/App.vue
```diff
@@ -172,6 +173,8 @@ import PlioAPIService from "@/services/API/Plio.js";
 import { mapActions, mapState, mapGetters } from "vuex";
 import { useToast } from "vue-toastification";
 
+const MOBILE_SCREEN_WIDTH_THRESHOLD = 500;
```
deepansh96, line 177: - Make the other value in `PlioListItem.vue` as TAB_SCREEN_WIDTH_THRESHOLD
- Also, make these constants and methods (`isMobileScreen` and `isTabScreen`) as methods imported from the Utilities module. And those modules will use the store value in turn.

deepansh96 (changes requested):

deepansh96 (approved):

PR #526, opened 2021-11-15
Cypress integration

### LINKED ISSUE #525: Integration & E2E testing
**Describe the task**
Integrate Cypress that performs Integration and E2E testing

rathorevaibhav:
Fixes #525

## Summary

1. Cypress integration
2. Please check Cypress.md on how to run cypress tests

## Test Plan

- [ ] Test Responsiveness
   - [ ] Laptop (1200px)
   - [ ] Tablet (760px)
   - [ ] Phone (320px)
- [ ] Cross-Browser Testing
   - [ ] Chrome
   - [ ] Firefox
- [ ] Local Language Support
- [x] Wrote tests
- [x] Tested locally
- [ ] Self-review
- [ ] Tested on staging
- [ ] Tested on production


src/main.js
```diff
@@ -61,6 +61,11 @@ const filterBeforeCreate = (toast, toasts) => {
   return toast;
 };
 
+// Cypress automatically sets window.Cypress by default
+if (window.Cypress) {
+  window.__store__ = store;
```
dalmia, line 66: it's not clear to me what this attribute means and why we need to set this variable when cypress is being used

cypress.json
```diff
@@ -0,0 +1,3 @@
+{
+  "baseUrl": "http://localhost:8080"
+}
```
deepansh96, line 3: Why isn't this a part of `cypress.env.example.json`?
And can we do cypress testing only locally?
We can't do E2E testing on staging or prod?

cypress/integration/home/home.spec.js
```diff
@@ -0,0 +1,9 @@
+describe("sees home", () => {
```
deepansh96, line 1: Any specific reason for keeping the file structure this way?
`cypress/integration/home...`?

As our unit tests are in this way
`./tests/unti/...`
so doesn't it makes sense to keep integration tests in the same way?
`./tests/integration/...`

docs/CYPRESS.md
```diff
@@ -0,0 +1,48 @@
...
+To configure your local setup for Cypress, only the following steps are needed:
+1. Set up the Plio Frontend, Plio Backend and Plio Analytics applications on your local Docker.
+2. Copy `cypress.env.example.json` and rename it as `cypress.env.json`
+3. Update the JSON object in `cypress.env.json` by setting the correct configurations
+
+### Cypress ENV variables
+Plio uses `cypress.env.json` to set up environment variables for the Cypress app. Below are the details of all the environment variables.
+
+#### `baseUrl`
+Base url for the frontend App. By default the value is `http://localhost:8080`.
```
deepansh96, line 30: As I've mentioned in the comment above, this `baseUrl` is in env.json and not cypress.env.json

src/pages/Login.vue
```diff
@@ -434,6 +434,7 @@ export default {
         let socialAuthToken = googleUser.getAuthResponse();
         UserAPIService.convertSocialAuthToken(socialAuthToken.access_token).then(
           (response) => {
+            console.log('response', response)
```
deepansh96, line 437: Will you remove this later on, before merging?

cypress/support/commands.js
```diff
@@ -0,0 +1,61 @@
...
+// https://on.cypress.io/custom-commands
+// ***********************************************
+//
+//
+// -- This is a parent command --
+// Cypress.Commands.add('login', (email, password) => { ... })
+//
+//
+// -- This is a child command --
+// Cypress.Commands.add('drag', { prevSubject: 'element'}, (subject, options) => { ... })
```
deepansh96, line 17:
```suggestion
// Cypress.Commands.add('drag', { prevSubject: 'element' }, (subject, options) => { ... })
```

src/main.js
deepansh96: This is just to add the vuex store into the global namespace -- as it's used by commands.js file

cypress/plugins/index.js
```diff
@@ -0,0 +1,22 @@
+/// <reference types="cypress" />
```
deepansh96, line 1: Didn't get the point of this file

dalmia (changes requested):

dalmia:
Screenshot 2021-11-17 at 3 09 04 PM

this security test is failing!

src/main.js
rathorevaibhav: Yes, this is the vuex store. We need it within Cypress also, hence we're registering the `store` with the global `window` object

cypress.json
rathorevaibhav: This is something that can't be overridden with cypress env as mentioned here.
https://docs.cypress.io/guides/guides/environment-variables#Setting

In case we need to run on a different base URL, we can control it via command line while running cypress
```
CYPRESS_BASE_URL=https://staging.app.com npx cypress open
```

cypress/integration/home/home.spec.js
rathorevaibhav: No specific reason. That's how cypress created directories when I installed it and I went with default. Let me know if you think there can be a different structure. We can easily control it.

docs/CYPRESS.md
rathorevaibhav: I've updated the docs accordingly.

src/pages/Login.vue
rathorevaibhav: Thanks for catching it. Removed!

cypress/plugins/index.js
rathorevaibhav: Yes. This file comes by default with cypress installation. We may require it in future when we start writing more deeper test cases.

docs/CYPRESS.md
```diff
@@ -0,0 +1,48 @@
+## Cypress
+This guide aims to share how Cypress is being used in Plio for performing E2E and integration tests.
+
+### Running the tests locally
+Run the following command to run the Cypress tests. Make sure the Plio Frontend, Plio Backend and Plio Analytics applications are running on your local Docker.
+```sh
+npx cypress open
```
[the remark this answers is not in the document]
rathorevaibhav: Yes, I think you are right. Just `npx cypress open` is working on my machine but I believe on CI or some other machine not having npm installed will throw an error.

Let me update.

cypress/integration/home/home.spec.js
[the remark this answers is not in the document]
rathorevaibhav: I'm not sure if I understood your comment. Isn't it already `home.spec.js`?

cypress/support/commands.js
```diff
@@ -0,0 +1,61 @@
...
+  cy.window().then((window) => {
+    cy.request({
+      method: "POST",
+      url: "https://www.googleapis.com/oauth2/v4/token",
+      body: {
+        grant_type: "refresh_token",
+        client_id: Cypress.env("google_auth").client_id,
+        client_secret: Cypress.env("google_auth").client_secret,
+        refresh_token: Cypress.env("google_auth").refresh_token,
+      },
```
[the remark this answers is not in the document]
rathorevaibhav: Yes, this one is similar to how we're maintaining the ENV files across our systems, however, this doesn't support nested env variables. And we'll have to install additional package (dotenv) to have it running.

The one I have implemented comes by default and supports nested variables, so I thought it's a bit faster and easier to set up.

Also, the `.env` file in the screenshot above needs to reside in `cypress/` folder. That means we will have two different env files, one in root folder and another one in cypress. The structure looked a tiny bit confusing.

cypress/support/commands.js
```diff
@@ -0,0 +1,61 @@
...
+          grant_type: "convert_token",
+          client_id: Cypress.env("plio_backend").client_id,
+          client_secret: Cypress.env("plio_backend").client_secret,
+          backend: "google-oauth2",
+          token: access_token,
+        },
+      }).then(({ body }) => {
+        window.__store__.dispatch("auth/setAccessToken", body);
+        cy.visit("/home");
+      });
```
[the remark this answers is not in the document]
rathorevaibhav: Not sure if I got your point but shouldn't the error help that something is not right?

docs/CYPRESS.md
```diff
@@ -0,0 +1,48 @@
+## Cypress
```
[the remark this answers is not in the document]
rathorevaibhav: Didn't get your point, please elaborate?

cypress.json
rathorevaibhav: We'll run cypress on CI. That I've started working on.

src/pages/Login.vue
```diff
@@ -14,7 +14,7 @@
           alt="Plio logo"
         />
 
-        <p class="mt-2 text-center text-md sm:text-xl font-bold text-primary">
+        <p class="mt-2 text-center text-md sm:text-xl font-bold text-primary" data-test="loginHeading">
```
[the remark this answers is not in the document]
rathorevaibhav: I've added the test case that uses this. Thanks for catching!

cypress.env.example.json
```diff
@@ -0,0 +1,12 @@
+{
+  "plio_backend": {
+    "convert_social_auth_token_url": "http://0.0.0.0:8001/auth/convert-token",
+    "client_id": "",
+    "client_secret": ""
+  },
+  "google_auth": {
```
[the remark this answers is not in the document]
rathorevaibhav: Done!

cypress/integration/home/home.spec.js
```diff
@@ -0,0 +1,9 @@
...
+  beforeEach(() => {
+    cy.loginByGoogleApi();
+  });
+
+  it("sees the plio list", () => {
```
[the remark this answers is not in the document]
rathorevaibhav: Yes, it's actually using my local setup and will use yours when you run locally.

To run cypress on CI, we'll have to set up the Plio apps first and then the cypress test cases will use it.

src/main.js
dalmia: My doubt was also regarding why are we using `window.__store__` instead of, say, `window.store`? Is there a convention?

cypress.json
dalmia: So, we'll be using both `cypress.json` and `cypress.env.json`? From the documentation page, it seemed like we should be using either one of those. Also, it's not clear to me how we are parameterising the baseUrl just yet for being used in CI. Should I wait for you to update the PR with the code required to perform CI for that?

cypress/integration/home/home.spec.js
dalmia: I agree with Deepansh here too. Let's keep the structure that Deepansh proposed.

docs/CYPRESS.md
```diff
@@ -23,14 +23,13 @@ Follow the links below one-by-one to retrieve required information from Google:
...
 2. Copy `cypress.env.example.json` and rename it as `cypress.env.json`
-3. Update the JSON object in `cypress.env.json` by setting the correct configurations
-
-### Cypress ENV variables
-Plio uses `cypress.env.json` to set up environment variables for the Cypress app. Below are the details of all the environment variables.
+3. Update the JSON object in `cypress.env.json` by setting the correct configurations - see below.
 
+#### Cypress ENV variables
+ Below are the details of all the environment variables.
 #### `google_auth.refresh_token`
```
dalmia, line 33: shouldn't all these headings be H5 now?

docs/CYPRESS.md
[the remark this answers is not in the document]
dalmia: Just flagging this again as I didn't see an update regarding this.

cypress/integration/home/home.spec.js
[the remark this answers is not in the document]
dalmia: Oh I meant the name used under `describe`:
```suggestion
describe("Home", () => {
```

cypress/support/commands.js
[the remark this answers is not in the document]
dalmia: as in, shouldn't there be a `catch` in case the POST request returned an error?

docs/CYPRESS.md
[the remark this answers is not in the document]
dalmia: As in, the details in this doc seem to be for local testing and not for testing on staging when the PR is created - just like we run the unit tests for example.

cypress/support/commands.js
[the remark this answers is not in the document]
rathorevaibhav: I don't think catching the error will be of any benefit as without the auth token from Google, we can't proceed further.

If there's an error, that means there's something wrong with Google Auth, which will notify us to check the Google Login on production as well.

cypress.json
rathorevaibhav: So Cypress.json will contain all common configurations that will be used in everyone's machine (including CI). It is similar to a file say `config.js`. Cypress env contains configurations specific to developer's machine and the values within can be different for everyone.

BaseURL is not supported in cypress env as per the documentation link I've shared above. They recommend to have a default base URL mentioned in cypress.json and in case need to be overridden, it can be done via the command (mentioned above also):
```
CYPRESS_BASE_URL=https://staging.app.com npx cypress open
```

rathorevaibhav: For CI, we may have to use the similar command instead of just running `npx cypress open` if the baseUrl is not the default one that we have in cypress.json

src/main.js
rathorevaibhav: There's no convention. I believe it's mostly to avoid conflict with existing functionality and other packages. `window.store` can be initialized or used by some other functionalities. A different name gives a bit more flexibility and chances of avoiding conflicts.

cypress/integration/home/home.spec.js
rathorevaibhav: Done.

PR #534, opened 2021-11-24
FIX: update 'watching' events instead of re-creating them everytime

### LINKED ISSUE #532: Increase upload interval time from 10 secs to 20 secs

### LINKED ISSUE #533: Update 'watching' events instead of creating everytime
Whenever a session is created, and the user watches a plio, we create new 'watching' events every few seconds. This has added up over the past few months and is now taking 80% of space in the Events table on our database.
This also led to an out of memory error in our data pipeline.

Instead of creating new watching events, just create **one** watching event for each session and keep updating it regularly.

deepansh96:
Fixes #532 
Fixes #533 

## Test Plan

- [ ] ~Test Responsiveness~
   - [ ] Laptop (1200px)
   - [ ] Tablet (760px)
   - [ ] Phone (320px)
- [ ] ~Cross-Browser Testing~
   - [ ] Chrome
   - [ ] Firefox
- [x] Wrote tests
- [x] Tested locally
- [x] Self-review
- [ ] Tested on staging
- [ ] Tested on production

### DIFF
--- a/src/pages/Embeds/Plio.vue
+++ b/src/pages/Embeds/Plio.vue
@@ -1064,13 +1064,12 @@ export default {
        * in preview mode
        */
       if (!this.hasSessionStarted || !this.isAuthenticated || this.previewMode) return;
-      var eventData = {
+      let response = await EventAPIService.createEvent({
         type: eventType,
         details: eventDetails,
         player_time: this.player.currentTime != null ? this.player.currentTime : 0,
         session: this.sessionDBId,
-      };
-      let response = await EventAPIService.createEvent(eventData);
+      });
       if (eventType == "watching") this.watchingEventDBId = response.id;
     },
     /**
@@ -1080,13 +1079,12 @@ export default {
      * @param {Object} eventDetails - details of the event
      */
     updateEvent(eventType, eventDBId, eventDetails = {}) {
-      let eventData = {
+      EventAPIService.updateEvent(eventDBId, {
         type: eventType,
         details: eventDetails,
         player_time: this.player.currentTime != null ? this.player.currentTime : 0,
         session: this.sessionDBId,
-      };
-      EventAPIService.updateEvent(eventDBId, eventData);
+      });
     },
     goFullscreen() {
       this.isFullscreen = true;


src/pages/Embeds/Plio.vue
```diff
@@ -1050,8 +1054,10 @@ export default {
     },
     /**
      * creates a new event
+     * @param {String} eventType - The type of event that needs to be logged
+     * @param {Object} eventDetails - An object with the details of the event
```
dalmia, line 1058:
```suggestion
     * @param {Object} eventDetails - details of the event
```

src/pages/Embeds/Plio.vue
```diff
@@ -726,8 +727,11 @@ export default {
       if (this.hasSessionStarted) {
         // update session data
         this.updateSession();
-        // create an event for the user watching the plio
-        this.createEvent("watching");
+        // if a 'watching' event exists for the current session, update that event
+        // else create a new event for the user watching the plio
```
dalmia, line 731:
```suggestion
        // else create a new event
```

dalmia, line 731:
```suggestion
        // else create a new event
```

src/pages/Embeds/Plio.vue
dalmia, line 1058:
```suggestion
     * @param {Object} eventDetails - details of the event
```

src/pages/Embeds/Plio.vue
```diff
@@ -1064,7 +1070,24 @@ export default {
...
       };
-      EventAPIService.createEvent(eventData);
+      let response = await EventAPIService.createEvent(eventData);
+      if (eventType == "watching") this.watchingEventDBId = response.id;
+    },
+    /**
+     * Updates an event
+     * @param {String} eventType - The type of event that needs to be logged
+     * @param {Number} eventDBId - The id of an event as pulled from the database which needs to be updated
+     * @param {Object} eventDetails - An object with the details of the event
```
dalmia, line 1080:
```suggestion
     * @param {Object} eventDetails - details of the event
```

src/pages/Embeds/Plio.vue
```diff
@@ -1064,7 +1070,24 @@ export default {
...
         session: this.sessionDBId,
       };
-      EventAPIService.createEvent(eventData);
+      let response = await EventAPIService.createEvent(eventData);
+      if (eventType == "watching") this.watchingEventDBId = response.id;
+    },
+    /**
+     * Updates an event
+     * @param {String} eventType - The type of event that needs to be logged
+     * @param {Number} eventDBId - The id of an event as pulled from the database which needs to be updated
```
dalmia, line 1079:
```suggestion
     * @param {Number} eventDBId - The id of the event instance which needs to be updated
```

src/pages/Embeds/Plio.vue
```diff
@@ -1064,7 +1070,24 @@ export default {
...
+     * @param {Object} eventDetails - An object with the details of the event
+     */
+    updateEvent(eventType, eventDBId, eventDetails = {}) {
+      if (!this.hasSessionStarted || !this.isAuthenticated || this.previewMode) return;
```
dalmia, line 1083: this is not needed as the function invoking this function already places these checks.

src/pages/Embeds/Plio.vue
```diff
@@ -1064,7 +1070,24 @@ export default {
...
+      let eventData = {
```
dalmia, line 1084: why create this variable?

src/services/API/Event.js
```diff
@@ -2,6 +2,10 @@ import { apiClient } from "@/services/API/RootClient.js";
 import { eventsEndpoint } from "@/services/API/Endpoints.js";
 
 export default {
+  /**
+   *
+   * @param {Object} eventData - The data that will be used as a payload in the POST request
```
dalmia, line 7:
```suggestion
   * @param {Object} eventData - details for the event to be created
```

src/services/API/Event.js
```diff
@@ -10,4 +14,12 @@ export default {
         return response.data;
       });
   },
+  /**
+   *
```
dalmia, line 18: description for the method?

src/services/API/Event.js
```diff
@@ -2,6 +2,10 @@ import { apiClient } from "@/services/API/RootClient.js";
 import { eventsEndpoint } from "@/services/API/Endpoints.js";
 
 export default {
+  /**
+   *
```
dalmia, line 6: description for the method?

src/services/API/Event.js
```diff
@@ -10,4 +14,12 @@ export default {
         return response.data;
       });
   },
+  /**
+   *
+   * @param {Number} eventId - The ID of the event that needs to be patched
```
dalmia, line 19:
```suggestion
   * @param {Number} eventId - The ID of the event that needs to be updated
```

src/services/API/Event.js
```diff
@@ -10,4 +14,12 @@ export default {
...
+   * @param {Object} eventData - The event will be patched with this data
```
dalmia, line 20:
```suggestion
   * @param {Object} eventData - The event will be updated with this data
```

src/services/API/Event.js
```diff
@@ -10,4 +14,12 @@ export default {
...
+   */
+  updateEvent(eventId, eventData) {
+    return apiClient().patch(eventsEndpoint + eventId, eventData);
```
dalmia, line 23: we are sending the entire event details object, no? then, put will be more appropriate no?

tests/unit/pages/embeds/Plio.spec.js
```diff
@@ -318,4 +319,96 @@ describe("Plio.vue", () => {
...
+    // using the fake response data
+    mockAxios.mockResponse(clonedeep(dummyPublishedPlio), mockAxios.queue()[0]);
+
+    // wait until the DOM updates after promises resolve
+    await flushPromises();
+    // assert that updateEvent method has been called with the correct params
+    expect(updateEvent).toHaveBeenCalledWith("watching", 1);
+  });
+
+  it("makes a PATCH call to update a watching event", () => {
```
dalmia, line 380:
```suggestion
  it("updates a watching event when updateEvent is called", () => {
```

src/pages/Embeds/Plio.vue
```diff
@@ -191,6 +191,7 @@ export default {
       isScorecardShown: false, // to show the scorecard or not
       plioTitle: "", // title of the plio
       isAspectRatioChecked: false, // whether the check for aspect ratio has been done
+      watchingEventDBId: null, // the DB id of the latest 'watching' event for a given session
```
dalmia, line 194: I think there is no other id it can be apart from the DB id - plus you've mentioned it already in the comments.
```suggestion
      watchingEventId: null, // the DB id of the latest 'watching' event for a given session
```

tests/unit/pages/embeds/Plio.spec.js
```diff
@@ -318,4 +319,96 @@ describe("Plio.vue", () => {
...
+    await store.dispatch("auth/setAccessToken", dummyAccessToken);
+
+    mount(Plio, {
+      shallow: true,
+      props: {
+        plioId: plioId,
+      },
+      data() {
+        return {
+          watchingEventDBId: 1,
```
dalmia, line 362: comment saying that this will simulate the case when watching event was already created

tests/unit/pages/embeds/Plio.spec.js
```diff
@@ -318,4 +319,96 @@ describe("Plio.vue", () => {
...
+    await flushPromises();
+
+    // resolve the `GET` request waiting in the queue (for receiving plio details)
+    // using the fake response data
+    mockAxios.mockResponse(clonedeep(dummyPublishedPlio), mockAxios.queue()[0]);
+
+    // wait until the DOM updates after promises resolve
+    await flushPromises();
+    // assert that updateEvent method has been called with the correct params
+    expect(updateEvent).toHaveBeenCalledWith("watching", 1);
```
dalmia, line 377: use a variable for the db id and pass it in both the mount and this assertion.

tests/unit/pages/embeds/Plio.spec.js
```diff
@@ -318,4 +319,96 @@ describe("Plio.vue", () => {
...
+    };
+
+    let eventType = "watching";
+    let eventDBId = 1;
+
+    // invoke the method using the local context object and appropriate params
+    Plio.methods.updateEvent.call(localThis, eventType, eventDBId);
+
+    // the updateEvent method inside EventAPIService would've been called
+    // A patch request would've been made with the correct payload
```
dalmia, line 404: we don't need to explain which request types to expect in the comments - we have not done that for any API request tests + it is fairly obvious from the code + it ties the comment to the implementation (which we'll have to remember to update if we update the method)

dalmia (changes requested):

src/pages/Embeds/Plio.vue
deepansh96: Just followed the style of the method `createEvent`
Also makes it a little more readable. Passing `eventData` to EventAPIService.udpateEvent()` feels more natural and readable.

If we'll get rid of this variable here, then we should do the same in `createEvent` method

src/services/API/Event.js
deepansh96: As discussed, skippping it as method name is obvious

src/services/API/Event.js
deepansh96: As discussed, skippping it as method name is obvious

src/services/API/Event.js
deepansh96: Updated it according to the method `createEvent` above.
Also, I think let's stick with `payload` term. It is generic enough and it makes sense in every API call.

src/pages/Embeds/Plio.vue
dalmia: haan makes sense - let's do that

tests/unit/pages/embeds/Plio.spec.js
```diff
@@ -352,15 +352,19 @@ describe("Plio.vue", () => {
     // set user
     await store.dispatch("auth/setAccessToken", dummyAccessToken);
 
+    // stimulating the case where a 'watching' event has already been created
```
dalmia, line 355:
```suggestion
    // simulating the case where a 'watching' event has already been created
```

dalmia (changes requested):

PR #542, opened 2021-11-26
FIX: removed primevue tooltip, used tippy tooltip

### LINKED ISSUE #541: Implement a non-buggy tooltip, remove the old primevue tooltip
As we all know, Primevue is super buggy
The tooltip they provided was also super buggy

Switched to a new, much better option!
[Tippy.js](https://atomiks.github.io/tippyjs/)

They use [Popper.js](https://popper.js.org/) under the hood.

Integrating it into vue as a global plugin was made easier by [this](https://vue-tippy.netlify.app/) wrapper.

deepansh96:
Fixes #541 

## Test Plan

- [x] Test Responsiveness
   - [ ] Laptop (1200px)
   - [ ] Tablet (760px)
   - [ ] Phone (320px)
- [x] Cross-Browser Testing
   - [ ] Chrome
   - [ ] Firefox
- [x] Local Language Support
- [x] Wrote tests
- [x] Tested locally
- [x] Self-review
- [x] Check for bundle size [here](https://bundlephobia.com/) if adding a package
- [ ] Tested on staging
- [ ] Tested on production


jest.init.js
```diff
@@ -6,7 +6,7 @@
 
 import { config } from "@vue/test-utils";
 import Toast from "vue-toastification";
-import Tooltip from "primevue/tooltip";
+import VueTippy from "vue-tippy";
```
dalmia, line 9: we can use a more readable name?
```suggestion
import VueTooltip from "vue-tippy";
```

src/components/Editor/ItemEditor.vue
```diff
@@ -47,27 +47,34 @@
...
-      ></icon-button>
+      <!-- adding an enclosing span to show tooltip when disabled -->
+      <span v-tooltip="deleteItemButtonTooltip" tabindex="0">
+        <icon-button
+          class="rounded-xl bg-red-600 hover:bg-red-700 w-8 h-8 shadow-lg px-2"
+          :iconConfig="deleteItemIconConfig"
+          @click="deleteSelectedItem"
+          :buttonClass="deleteItemButtonClass"
+          :disabled="isInteractionDisabled"
+          v-tooltip="deleteItemButtonTooltip"
```
dalmia, line 73: can remove this then?

src/components/Editor/ItemEditor.vue
```diff
@@ -47,27 +47,34 @@
...
-        data-test="addItem"
-      ></icon-button>
+      <!-- adding an enclosing span to show tooltip when disabled -->
+      <span v-tooltip="addItemButtonTooltip" tabindex="0">
+        <icon-button
+          class="rounded-xl w-8 h-8 px-2 hidden bp-360:block"
+          :iconConfig="addItemIconConfig"
+          :buttonClass="addItemButtonClass"
+          @click="removeSelectedItemIndex"
+          v-tooltip="addItemButtonTooltip"
```
dalmia, line 57: can remove this then? (do check for everywhere that you've added span if there is already a v-tooltip added and remove it)

src/components/Editor/ItemEditor.vue
```diff
@@ -572,13 +581,14 @@ export default {
       // tooltip text for delete item button
       // itemType is just "question" right now - parametrize when more types are supported
```
dalmia, line 583: fix the docstring too?

src/main.js
```diff
@@ -5,12 +5,11 @@ import store from "./store";
 
 import i18n from "./services/Localisation/i18n.js";
 import InlineSvg from "vue-inline-svg";
-import PrimeVue from "primevue/config";
-import Tooltip from "primevue/tooltip";
 import Toast from "vue-toastification";
 import VueProgressBar from "@aacassandra/vue3-progressbar";
 import VueClickAway from "vue3-click-away";
 import mixpanel from "mixpanel-browser";
+import VueTippy from "vue-tippy";
```
dalmia, line 12: more readable import as before

src/main.js
```diff
@@ -93,12 +90,17 @@ if (
 
 app.component("inline-svg", InlineSvg);
 app.use(i18n);
-app.use(PrimeVue, { ripple: true });
+app.use(VueTippy, {
+  // https://vue-tippy.netlify.app/installation
+  directive: "tooltip", // => use as <button v-tippy="your text"></button>
```
dalmia, line 95: this, right?
```suggestion
  directive: "tooltip", // => use as <button v-tooltip="your text"></button>
```

src/main.js
```diff
@@ -93,12 +90,17 @@ if (
 
 app.component("inline-svg", InlineSvg);
 app.use(i18n);
-app.use(PrimeVue, { ripple: true });
+app.use(VueTippy, {
+  // https://vue-tippy.netlify.app/installation
```
dalmia, line 94: why is this added?

src/components/UI/Text/TimeInput.vue
```diff
@@ -46,7 +46,7 @@
         :maxLength="2"
         :boxStyling="[defaultBoxClass, secondInputInvalidClass]"
         :isDisabled="isDisabled"
-        v-tooltip.bottom="{ value: disabledInputTooltip, disabled: !isDisabled }"
+        v-tooltip="disabledInputTooltip"
```
dalmia, line 49: seems like the name for the tooltip variable in this file is not quite accurate.

src/main.js
```diff
@@ -93,12 +90,17 @@ if (
...
+  directive: "tooltip", // => use as <button v-tippy="your text"></button>
+  defaultProps: {
+    placement: "auto",
```
dalmia, line 97: Screenshot 2021-11-27 at 3 47 59 AM
is `auto` the right choice? for example, this doesn't look that good to me in terms of positioning.

src/main.js
```diff
@@ -93,12 +90,17 @@ if (
 
 app.component("inline-svg", InlineSvg);
 app.use(i18n);
-app.use(PrimeVue, { ripple: true });
+app.use(VueTippy, {
```
dalmia, line 93: Also, since we now have a good implementation of tooltip - do you want to add a tooltip to all the important parts? can we done in a separate PR too. For example, on the editor, the publish button has a tooltip but the home button on its side does not. Ideally, the marker can have a longer tooltip saying that clicking on the marker will select that item and that timestamp can be changed by dragging it, etc.

dalmia (changes requested):

src/main.js
deepansh96: for reference on how to import this plugin

src/services/Functional/Utilities.js
```diff
@@ -173,3 +173,44 @@ export function isScreenPortrait() {
...
+    .substr(11, 8);
+
+  timestampObject["hour"] = parseInt(isoTime.split(":")[0]);
+  timestampObject["minute"] = parseInt(isoTime.split(":")[1]);
+  timestampObject["second"] = parseInt(isoTime.split(":")[2]);
+  timestampObject["millisecond"] = 0;
+
+  if (Math.floor(timeInSeconds) < timeInSeconds) {
+    var ms = parseInt(String(timeInSeconds).split(".")[1].padEnd(3, "0"));
+    timestampObject["millisecond"] = ms;
```
dalmia, line 199:
```suggestion
    timestampObject["millisecond"] = parseInt(String(timeInSeconds).split(".")[1].padEnd(3, "0"));
```

src/services/Functional/Utilities.js
```diff
@@ -173,3 +173,44 @@ export function isScreenPortrait() {
...
+/**
+ * Converts a timestamp in ISO format to seconds
+ * @param {Object} timeInISO - An object containing the ISO time as it's keys
+ * @returns {Number} - The converted time in seconds
+ */
+export function convertISOTimeToSeconds(timeInISO) {
+  var hour = parseInt(timeInISO.hour) || 0;
+  var minute = parseInt(timeInISO.minute) || 0;
+  var second = parseInt(timeInISO.second) || 0;
+  var millisecond = parseInt(timeInISO.millisecond) || 0;
```
dalmia, line 214: use let

src/components/UI/Slider/SliderWithMarkers.vue
```diff
@@ -83,6 +94,37 @@ export default {
...
+     * Update the content and positioning of a marker tooltip
+     * @param {Number} markerIndex - The index of the marker
+     */
+    updateMarkerTooltip(markerIndex) {
+      if (markerIndex == null) {
+        // don't show anything
+        this.markerTooltipContent = null;
+      } else {
+        // convert the time in seconds to ISO time
+        let timeInSeconds = this.localMarkerPositions[markerIndex];
```
dalmia, line 107: no need to create this variable no?

src/components/UI/Slider/SliderWithMarkers.vue
```diff
@@ -83,6 +94,37 @@ export default {
...
+        let ISOTimeObject = convertSecondsToISOTime(timeInSeconds);
+        this.markerTooltipContent =
+          ISOTimeObject.minute +
+          ":" +
+          ISOTimeObject.second +
+          ":" +
+          ISOTimeObject.millisecond;
```
dalmia, line 114: a cleaner one might this:

```suggestion
        this.markerTooltipContent = `${ISOTimeObject.minute}:${ISOTimeObject.second}:${ISOTimeObject.millisecond}`
```

and then, can apply this for the string concatenations below

dalmia (changes requested):

tests/unit/pages/Editor.spec.js
```diff
@@ -1482,7 +1482,7 @@ describe("Editor.vue", () => {
     expect(wrapper.vm.dialogAction).toBe("deleteOption");
     expect(wrapper.vm.isDialogBoxShown).toBeTruthy();
 
-    expect(wrapper.vm.optionIndexToDelete).toBe(0);
+    expect(wrapper.vm.optionIndexToDelete).toBe(1);
```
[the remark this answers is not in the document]
deepansh96: Not sure why it was not `1` before.
If you see above, the option for which the delete button is being clicked is the last option
So option index to delete should be 1 only

tests/integration/integration/home/home.spec.js
```diff
@@ -4,6 +4,15 @@ describe("Home", () => {
   });
 
   it("the plio list is loaded", () => {
-    cy.get('[data-test="table"]').should("be.visible");
+    cy.intercept({
+      method: "GET",
+      url: "http://0.0.0.0:8001/api/v1/plios/list_uuid/?page=1",
```
dalmia, line 9: the baseUrl is already defined in the config. so use that, no?

dalmia: + you can import the endpoint from the `endpoints.js` file too.

src/components/UI/Text/TimeInput.vue
```diff
@@ -86,12 +86,9 @@ import InputText from "@/components/UI/Text/InputText.vue";
 
 export default {
   data() {
-    // don't show the hour part if the value is 0
-    let showHour = true;
-    if (this.timeObject.hour == 0) showHour = false;
     return {
       defaultConfig: {
-        showHour: showHour,
+        showHour: this.timeObject.hour != 0, // don't show the hour part if the value is 0
```
dalmia, line 91: have you tested it with a video that is more than an hour long just in case?

tests/unit/components/UI/Text/TimeInput.spec.js
```diff
@@ -5,11 +5,9 @@ describe("TimeInput.vue", () => {
...
   it("should render with default values", () => {
     expect(wrapper).toBeTruthy();
-    if (wrapper.vm.localConfig.showHour)
-      expect(
-        wrapper.find('[data-test="hour"]').find('[data-test="input"]').element
-          .value
-      ).toBe("0");
+    expect(wrapper.find('[data-test="hour"]').exists()).toBe(
+      wrapper.vm.localConfig.showHour
+    );
```
dalmia, line 10: I don't think this is a valid check - what if showHour was being set incorrectly and hence, the div will incorrectly not be shown? We won't be able to catch it with this test, no?
Ideally, there should be two tests - for both when the hour one is shown and when it is not - and the equality should be checked with `true` or `false` directly.

src/components/UI/Slider/SliderWithMarkers.vue
```diff
@@ -104,20 +104,15 @@ export default {
...
         // Only show the hour value if it's available
-        if (ISOTimeObject.hour != 0) {
-          this.markerTooltipContent =
-            ISOTimeObject.hour + ":" + this.markerTooltipContent;
-        }
+        if (ISOTimeObject.hour != 0)
+          this.markerTooltipContent = `${ISOTimeObject.hour}:${this.markerTooltipContent}`;
+
         // adjust the position of the tooltip's start to take into account the marker width
         this.markerTooltipPosition =
```
dalmia, line 117: apply the same string concatenation logic here
for better readability, can create a variable of the computation below first.

src/assets/locales/en.js
```diff
@@ -444,13 +449,35 @@ export default {
...
       },
+      home: "Go to home",
+      preview:
+        "Click to preview the plio and see how it will look like to your viewers",
+      copy_draft_link:
+        "Copy a link of this draft plio to share it with your teammates",
+      analyze: "Analyse the data collected from the viewers of this plio",
+      share_plio: "Share this plio with your viewers",
+      play_plio: "Try out the created plio yourself",
+      embed_plio: "Embed this plio as an iframe",
```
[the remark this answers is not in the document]
dalmia: haan makes sense - I think we should just remove the `iframe` bit for those who will be familiar with the idea of embedding but not with `iframe` (and hence, might feel confused)

tests/unit/pages/Editor.spec.js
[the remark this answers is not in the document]
dalmia: the last option is index = 3 no?
Screenshot 2021-12-02 at 4 14 11 PM

src/components/UI/Slider/SliderWithMarkers.vue
```diff
@@ -114,10 +114,9 @@ export default {
           this.markerTooltipContent = `${ISOTimeObject.hour}:${this.markerTooltipContent}`;
 
         // adjust the position of the tooltip's start to take into account the marker width
-        this.markerTooltipPosition =
-          "left: " +
-          (this.markerRelativePositions[markerIndex] - this.markerWidthPercent / 2) +
-          "%";
+        let computedPercentPosition =
```
dalmia, line 117:
```suggestion
        let markerTooltipPercentPosition =
```

src/components/UI/Slider/SliderWithMarkers.vue
```diff
@@ -114,10 +114,9 @@ export default {
...
+          this.markerRelativePositions[markerIndex] - this.markerWidthPercent / 2;
+        this.markerTooltipPosition = `left: ${computedPercentPosition}%`;
```
dalmia, line 119: the name of this should ideally be `markerTooltipStyle`, no?

src/services/Functional/Utilities.js
```diff
@@ -184,18 +184,32 @@ export function convertSecondsToISOTime(timeInSeconds) {
...
   // https://stackoverflow.com/questions/1322732/convert-seconds-to-hh-mm-ss-with-javascript
 
-  let timestampObject = {};
+  let timestampObject = {
+    hour: null,
+    minute: null,
+    second: null,
+    millisecond: null,
+    /**
+     * Converts the timepart from a Number to a String and pads it with zeros accordingly
```
dalmia, line 193:
```suggestion
     * Converts the time part from a Number to a String and pads it with zeros accordingly
```

src/services/Functional/Utilities.js
```diff
@@ -184,18 +184,32 @@ export function convertSecondsToISOTime(timeInSeconds) {
...
+     * @param {String} timepart - "hour", "minute", "second" or "millisecond"
+     * @returns {String} - A string padded on the start with zeroes depending on the timepart
+     */
+    getAsString(timepart) {
```
dalmia, line 197: timeComponent

tests/unit/components/Collections/ListItems/PlioListItem.spec.js
```diff
@@ -415,27 +415,6 @@ describe("PlioListItem.vue", () => {
...
-
-    wrapper = mount(PlioListItem);
-
-    // click the option dropdown
-    await wrapper
-      .get('[data-test="optionDropdown"]')
-      .get('[data-test="toggleButton"]')
-      .trigger("click");
-
-    // there should be 7 buttons - edit, play, share, embed, duplicate, delete, analyse
```
dalmia, line 438: this comment should be retained

tests/unit/components/Collections/ListItems/PlioListItem.spec.js
```diff
@@ -415,27 +415,6 @@ describe("PlioListItem.vue", () => {
       .trigger("click");
 
     // there should be 6 buttons - edit, play, share, embed, duplicate, delete
```
dalmia, line 417: this comment should be deleted instead

PR #545, opened 2021-11-29
Feature/cypress ci

rathorevaibhav:
## Summary

1. CI for cypress
2. Added Cypress badge in readme. Will start showing status after merge
3. Documentation update for Cypress CI
4. Cypress ENV variables flattened as nested fields aren't supported when passing variables within CI. See [this link](https://docs.cypress.io/guides/guides/environment-variables#Export-cypress-env-variables-from-the-command-line)
5. Cypress Dashboard project id added for Plio. This helps in integration Cypress Bot and saving recordings from CI runs to Cypress Dashboard.
6. Awaiting Vue store dispatch action to not proceed unless user and auth details have been stored in vuex

## Test Plan

- [ ] Test Responsiveness
   - [ ] Laptop (1200px)
   - [ ] Tablet (760px)
   - [ ] Phone (320px)
- [ ] Cross-Browser Testing
   - [ ] Chrome
   - [ ] Firefox
   - [ ] Safari (use BrowserStack)
- [ ] Local Language Support
- [ ] Hygiene and Housekeeping
   - [ ] Self-review
   - [ ] Comments have been added appropriately
   - [ ] Check for bundle size [here](https://bundlephobia.com/) if adding a package
   - [ ] Added relevant details like Labels/Projects/Milestones etc.
- [ ] Testing
   - [ ] Wrote tests
   - [ ] Tested locally
   - [ ] Tested on staging
   - [ ] Tested on an actual physical phone
   - [ ] Tested on production
- [ ] Lighthouse Checks
   - [ ] Images have `alt` attributes
   - [ ] Any `` tags have `width` and `height` specified
   - [ ] Any `target="_blank"` links have `rel="noopener"`
   - [ ] Only SVGs are used as images. If PNGs are used, their size has been optimised.
   - [ ] Any SVGs without text have their `aria-label` attributes set

PR #549, opened 2021-12-03
Feature/cypress coverage

rathorevaibhav:
**Requires #545 to be merged before**

## Summary

1. Code coverage for cypress
2. CI setup with coveralls

## Test Plan

- [ ] Test Responsiveness
   - [ ] Laptop (1200px)
   - [ ] Tablet (760px)
   - [ ] Phone (320px)
- [ ] Cross-Browser Testing
   - [ ] Chrome
   - [ ] Firefox
- [ ] Local Language Support
   - [ ] Self-review
- [ ] Testing
   - [ ] Wrote tests
   - [ ] Tested locally
   - [ ] Tested on staging
   - [ ] Tested on production
- [ ] Lighthouse Checks


.github/workflows/ci.yml
```diff
@@ -140,6 +149,12 @@ jobs:
           # run cypress test cases
           npx cypress run --record --key ${{ secrets.CYPRESS_RECORD_KEY }}
 
+      - name: Coveralls
+        uses: coverallsapp/github-action@master
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
```
dalmia, line 155: is this a new environment variable?

rathorevaibhav: nope. this is built-in with GitHub actions. Coveralls uses this to find the integrated github repository at it's end and upload the codecov into the right project.

tests/integration/plugins/index.js
```diff
@@ -17,6 +17,11 @@
  */
 // eslint-disable-next-line no-unused-vars
 module.exports = (on, config) => {
+  require("@cypress/code-coverage/task.js")(on, config);
   // `on` is used to hook into various events Cypress emits
   // `config` is the resolved Cypress config
+
+  // It's IMPORTANT to return the config object
+  // with any changed environment variables
+  return config;
```
deepansh96, line 26: I did not understand why the need for this came up now, like, how was it that everything was working before without returning the config object?

.nycrc
```diff
@@ -0,0 +1,3 @@
+{
+  "extension": [".js", ".vue"]
+}
```
deepansh96, line 3: Didn't understand why this is needed!

.github/workflows/ci.yml
```diff
@@ -129,6 +129,15 @@ jobs:
         run: |
           cd projects/plio-frontend/
 
+          # delete the node_modules created by the docker
+          echo 'deleting node modules'
+          rm -rf node_modules/
```
deepansh96, line 134: When building the docker image, it does npm install and creates node_modules in it's own separate memory right? It doesn't create a node_modules in our app's folder no?

Maybe I'm confused

deepansh96 (approved):

PR #551, opened 2021-12-06
CI cleanups

rathorevaibhav:
## Summary

1. Removed duplicate variables
2. Removed GitHub environment. Directly picking from organization env variables

## Test Plan

- [ ] Test Responsiveness
   - [ ] Laptop (1200px)
   - [ ] Tablet (760px)
   - [ ] Phone (320px)
- [ ] Cross-Browser Testing
   - [ ] Chrome
   - [ ] Firefox
- [ ] Local Language Support
   - [ ] Self-review
- [ ] Testing
   - [ ] Wrote tests
   - [ ] Tested locally
   - [ ] Tested on staging
   - [ ] Tested on production
- [ ] Lighthouse Checks

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,11 +48,11 @@ jobs:
           cd plio-frontend/
 
           # check branch and switch to branch if exists
-          if [ `git branch --list --remote origin/${{ github.ref_name }}` ]
+          if [ `git branch --list --remote origin/${{ github.ref }}` ]
           then
-            echo "Switching to branch ${{ github.ref_name }}."
-            git checkout ${{ github.ref_name }}
-            git pull origin ${{ github.ref_name }}
+            echo "Switching to branch ${{ github.ref }}."
+            git checkout ${{ github.ref }}
+            git pull origin ${{ github.ref }}
           else
             echo "Branch not found. Going with default branch."
           fi
@@ -76,11 +76,11 @@ jobs:
           cd plio-backend/
 
           # check branch and switch to branch if exists
-          if [ `git branch --list --remote origin/${{ github.ref_name }}` ]
+          if [ `git branch --list --remote origin/${{ github.ref }}` ]
           then
-            echo "Switching to branch ${{ github.ref_name }}."
-            git checkout ${{ github.ref_name }}
-            git pull origin ${{ github.ref_name }}
+            echo "Switching to branch ${{ github.ref }}."
+            git checkout ${{ github.ref }}
+            git pull origin ${{ github.ref }}
           else
             echo "Branch not found. Going with default branch."
           fi
@@ -107,11 +107,11 @@ jobs:
           cd plio-analytics/
 
           # check branch and switch to branch if exists
-          if [ `git branch --list --remote origin/${{ github.ref_name }}` ]
+          if [ `git branch --list --remote origin/${{ github.ref }}` ]
           then
-            echo "Switching to branch ${{ github.ref_name }}."
-            git checkout ${{ github.ref_name }}
-            git pull origin ${{ github.ref_name }}
+            echo "Switching to branch ${{ github.ref }}."
+            git checkout ${{ github.ref }}
+            git pull origin ${{ github.ref }}
           else
             echo "Branch not found. Going with default branch."
           fi
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,11 +48,11 @@ jobs:
           cd plio-frontend/
 
           # check branch and switch to branch if exists
-          if [ `git branch --list --remote origin/${{ github.ref }}` ]
+          if [ `git branch --list --remote origin/${{ github.head_ref }}` ]
           then
-            echo "Switching to branch ${{ github.ref }}."
-            git checkout ${{ github.ref }}
-            git pull origin ${{ github.ref }}
+            echo "Switching to branch ${{ github.head_ref }}."
+            git checkout ${{ github.head_ref }}
+            git pull origin ${{ github.head_ref }}
           else
             echo "Branch not found. Going with default branch."
           fi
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -90,6 +106,16 @@ jobs:
...
+          # check branch and switch to branch if exists
+          if [ `git branch --list --remote origin/${{ github.head_ref }}` ]
+          then
+            echo "Switching to branch ${{ github.head_ref }}."
+            git checkout ${{ github.head_ref }}
+            git pull origin ${{ github.head_ref }}
+          else
+            echo "Branch not found. Going with default branch."
+          fi
+
```
dalmia, line 118: same comment on reusing the output of this step as mentioned in [avantifellows/plio-analytics#26](https://github.com/avantifellows/plio-analytics/pull/26)

.github/workflows/ci.yml
```diff
@@ -51,13 +47,23 @@ jobs:
           git clone https://github.com/avantifellows/plio-frontend/
           cd plio-frontend/
 
+          # check branch and switch to branch if exists
+          if [ `git branch --list --remote origin/${{ github.head_ref }}` ]
```
deepansh96, line 51: Screenshot 2021-12-07 at 12 54 03 PM

Here, the `head_ref` will only be available if this workflow is being run on a pull request. What happens when we merge to prod, will the cypress CI runs won't happen?

It's another story if we don't want them to run when merging branches.

deepansh96: I'm guessing during pushing to main/prod, as the `head_ref` will not be available, the code will just pull the default branch, right?

.github/workflows/ci.yml
```diff
@@ -51,13 +47,23 @@ jobs:
...
           # create the env file
           cp .env.example .env
 
           # add env secrets
-          echo 'VUE_APP_GOOGLE_CLIENT_ID=${{ secrets.VUE_APP_GOOGLE_CLIENT_ID }}' >> .env
-          echo 'VUE_APP_BACKEND_API_CLIENT_ID=${{ secrets.VUE_APP_BACKEND_API_CLIENT_ID }}' >> .env
-          echo 'VUE_APP_BACKEND_API_CLIENT_SECRET=${{ secrets.VUE_APP_BACKEND_API_CLIENT_SECRET }}' >> .env
+          echo 'VUE_APP_GOOGLE_CLIENT_ID=${{ secrets.GOOGLE_OAUTH2_CLIENT_ID }}' >> .env
+          echo 'VUE_APP_BACKEND_API_CLIENT_ID=${{ secrets.DJANGO_DEFAULT_OAUTH2_CLIENT_ID }}' >> .env
+          echo 'VUE_APP_BACKEND_API_CLIENT_SECRET=${{ secrets.DJANGO_DEFAULT_OAUTH2_CLIENT_SECRET }}' >> .env
```
deepansh96, line 66: Were these changed in the github secrets as well? And why was this change needed?

deepansh96 (approved):

.github/workflows/ci.yml
rathorevaibhav: That's a good point. I just checked and it seems to be returning empty value in case of direct push to the branch (without a pull request).

Since because of the empty value, the if condition will be false as it won't find a valid remote branch, hence it will go with default, which in our case is master. So I believe you're right that it will still run successfully.

.github/workflows/ci.yml
rathorevaibhav: @deepansh96 I've updated documentation for Cypress CI. Please have a look.

docs/CYPRESS.md
```diff
@@ -66,27 +66,28 @@ The Cypress Dashboard is a web interface that allows you to run and debug your C
...
-      2. `VUE_APP_BACKEND_API_CLIENT_SECRET`
-      3. `VUE_APP_GOOGLE_CLIENT_ID`
-   5. Plio Backend (Django) - Refer [Plio Backend ENV guide](https://github.com/avantifellows/plio-backend/blob/master/docs/ENV.md)
-      1. `DJANGO_DEFAULT_OAUTH2_CLIENT_ID`
-      2. `DJANGO_DEFAULT_OAUTH2_CLIENT_SECRET`
-      3. `DJANGO_DEFAULT_OAUTH2_CLIENT_SETUP`
-      4. `DJANGO_GOOGLE_OAUTH2_CLIENT_ID`
-      5. `DJANGO_GOOGLE_OAUTH2_CLIENT_SECRET`
-      6. `DJANGO_SECRET_KEY`
+The job `integration-tests` inside the GitHub workflow file [ci.yml](../.github/workflows/ci.yml) is used to run the integration tests in GitHub Actions. Since Cypress runs in multiple Plio repositories, it's optimal to go with organization secrets that can be shared across several repositories. This saves time and effort to create env variables in every repository.
```
deepansh96, line 69:
```suggestion
The job `integration-tests` inside the GitHub workflow file [ci.yml](../.github/workflows/ci.yml) is used to run the integration tests in GitHub Actions. Since Cypress runs require access to multiple Plio repositories, it's optimal to go with organization secrets that can be shared across several repositories. This saves time and effort to create env variables in every repository.
```

deepansh96 (changes requested):

docs/CYPRESS.md
dalmia, line 69:
```suggestion
The job `integration-tests` inside the GitHub workflow file [ci.yml](../.github/workflows/ci.yml) is used to run the integration tests in GitHub Actions. Since Cypress runs require setting up multiple Plio repositories, it's optimal to go with organization secrets that can be shared across several repositories. This saves time and effort to create env variables in every repository.
```

dalmia:
Screenshot 2021-12-07 at 3 02 47 PM

PR #553, opened 2021-12-08
Fixes the position of show question button on the player to change according to player size

### LINKED ISSUE #552: Setting the position of show question button has a lot of bugs

### LINKED ISSUE #554: Playing the video when "show question" button is active would keep showing the button for some time

dalmia:
Fixes #552 
Fixes #554 

## Summary
- instead of using the coordinates of the minimize button to set the position of the maximise button, creating a div already with the position of the maximize button predetermined and just toggling the visibility of that div based on whether the modal is maximized or minimized
- had to add the transition CSS in `Plio.vue` too or else the transition wouldn't happen when you visit `/play`. but the transition did happen when we opened the plio as a preview in the editor.
- if one played the video when`Show Question` button was active, it would take some time for `currentItemIndex` to be set to null. But in the meanwhile, `itemType` was already set to null and then, the text for `Show Question` becomes invalid. Setting the `currentItemIndex` to null right after the video is played resolves this issue

## Test Plan

- [x] Test Responsiveness
   - [ ] Laptop (1200px)
   - [ ] Tablet (760px)
   - [ ] Phone (320px)
- [x] Cross-Browser Testing
   - [ ] Chrome
   - [ ] Firefox
   - [ ] Self-review
- [ ] Testing
   - [ ] ~Wrote tests~
   - [x] Tested locally
   - [ ] Tested on staging
   - [ ] Tested on production
- [ ] Lighthouse Checks


.gitignore
```diff
@@ -31,3 +31,4 @@ package-lock.json
 
 # cypress
 cypress.env.json
+tests/integration/videos/
```
deepansh96, line 34: Is this needed?

src/pages/Editor.vue
```diff
@@ -106,39 +106,43 @@
               </div>
             </div>
             <div v-else data-test="videoPreview">
-              <div class="relative">
+              <div class="flex relative">
```
deepansh96, line 109: Why flex?
maximize button will be on top of the video player right? then why do we need to put them in a flexbox?

PR #556, opened 2021-12-08
Uses raw total count of plios to determine whether table should be shown or not

### LINKED ISSUE #555: Deleting last plio from a search result incorrectly removes table

dalmia:
Fixes #555 

## Summary
- earlier, we were using the `count` returned in the paginated response to determine if the table should be shown
- however, since on the backend, we were applying pagination on the search-filtered results, the count returns 0 and hence, the table was being removed when we deleting the last plio of a result search even though there were other plios not matching the search result
- sending another value from the backend now on the raw, unfiltered count of plios and using that to determine whether to show the table

## Test Plan

- [ ] ~Test Responsiveness~
   - [ ] Laptop (1200px)
   - [ ] Tablet (760px)
   - [ ] Phone (320px)
- [ ] ~Cross-Browser Testing~
   - [ ] Chrome
   - [ ] Firefox
   - [ ] Self-review
- [ ] Testing
   - [x] Wrote tests
   - [x] Tested locally
   - [ ] Tested on staging
   - [ ] Tested on production
- [ ] ~Lighthouse Checks~


deepansh96:
Looks good just one thing I didn't understand are those videos that have been committed along with the PR in the `integration/videos` path?

PR #603, opened 2022-01-27
Phone number digits can not be entered more than 10 and 0 input invalid

### LINKED ISSUE #602: Phone number input is exceeding the maxLength specified
**Describe the bug**

The length of phone number while entering at login page is exceeding the maxLength specified. 
The files to look at: Login.vue and InputNumber.vue .

**To Reproduce**
1. Go to 'login-Plio'
2. Enter more than 10 digits for the phone number

**Expected behavior**
It should ideally not let user enter more than 10 digit .

**Screenshots**
Capture

Salonichoudhary07:
Fixes #602 

## Summary

The `localvalue.length`  was returning undefined as it was not being converted into a string.
`.toString` function is added to the `localValue` in `keypress` function in `InputNumber.vue`.
`type` of `<input` is changed to `type`  which is a prop and has `default:"tel"` as there were many trailing zeros and the eventual phone number length was going way beyond 10 and it was still marked valid .
`If` condition is added in `keypress` so that any character is not accepted as input. 

## Test Plan

- [ ] ~Test Responsiveness~
   - [ ] Laptop (1200px)
   - [ ] Tablet (760px)
   - [ ] Phone (320px)
- [ ] ~Cross-Browser Testing~
   - [ ] Chrome
   - [ ] Firefox
   - [x] Self-review
       - [ ] update `docs/ENV.md`
       - [ ] update DockerFile
- [ ] Testing
   - [ ] ~Wrote tests~
   - [x] Tested locally
   - [ ] Tested on staging
   - [ ] Tested on production
- [ ] ~Lighthouse Checks~

### DIFF
--- a/src/components/UI/Text/InputNumber.vue
+++ b/src/components/UI/Text/InputNumber.vue
@@ -152,7 +152,7 @@ export default {
     localValue: {
       // local copy of the value prop
       get() {
-        return this.value.toString();
+        return this.value;
       },
       set(localValue) {
         this.$emit("update:value", localValue);
@@ -257,7 +257,7 @@ export default {
     },
     keyPress(event) {
       // invoked when a key is pressed
-      if (this.maxLength != null && this.localValue.length == this.maxLength) {
+      if (this.maxLength != null && this.localValue.toString().length == this.maxLength) {
         event.preventDefault();
         return;
       }
--- a/src/components/UI/Text/InputNumber.vue
+++ b/src/components/UI/Text/InputNumber.vue
@@ -53,7 +53,9 @@
         :class="[inputAreaClass, boxStyling]"
         :min="min"
         :max="max"
-        type="type"
+        type="tel"
+        inputmode="numeric"
+        pattern="[0-9]*"
         autocomplete="off"
         data-test="input"
         :disabled="isDisabled"
@@ -106,11 +108,6 @@ export default {
       default: "",
       type: [String, Number],
     },
-    /** the type of the input in the input box */
-    type: { 
-      default: 'tel' ,
-      type: String, 
-    },
     /** minimum value acceptable in a number textbox */
     min: {
       default: 0,
@@ -160,7 +157,7 @@ export default {
         return this.value;
       },
       set(localValue) {
-        this.$emit("update:value", localValue);
+       this.$emit("update:value", localValue);
       },
     },
     isTitlePresent() {
@@ -262,7 +259,12 @@ export default {
     },
     keyPress(event) {
       // invoked when a key is pressed
-      if (this.maxLength != null && this.localValue.length == this.maxLength) {
+      event = event || window.event;
+      var charCode = (typeof event.which == "undefined") ? event.keyCode : event.which;
+      var charStr = String.fromCharCode(charCode);
+      if (!charStr.match(/^[0-9]+$/))
+      event.preventDefault();
+      if (this.maxLength != null && this.localValue.toString().length == this.maxLength) {
         event.preventDefault();
         return;
       }
--- a/src/components/UI/Text/InputNumber.vue
+++ b/src/components/UI/Text/InputNumber.vue
@@ -53,9 +53,7 @@
         :class="[inputAreaClass, boxStyling]"
         :min="min"
         :max="max"
-        type="tel"
-        inputmode="numeric"
-        pattern="[0-9]*"
+        type="type"
         autocomplete="off"
         data-test="input"
         :disabled="isDisabled"
@@ -107,6 +105,11 @@ export default {
     value: {
       default: "",
       type: [String, Number],
+    },
+    /** the type of the input in the input box */
+    type: { 
+      default: 'tel' ,
+      type: String, 
     },
     /** minimum value acceptable in a number textbox */
     min: {
--- a/src/components/UI/Text/InputNumber.vue
+++ b/src/components/UI/Text/InputNumber.vue
@@ -53,7 +53,7 @@
         :class="[inputAreaClass, boxStyling]"
         :min="min"
         :max="max"
-        type="type"
+        :type="type"
         autocomplete="off"
         data-test="input"
         :disabled="isDisabled"
@@ -107,9 +107,9 @@ export default {
       type: [String, Number],
     },
     /** the type of the input in the input box */
-    type: { 
-      default: 'tel' ,
-      type: String, 
+    type: {
+      default: "tel",
+      type: String,
     },
     /** minimum value acceptable in a number textbox */
     min: {
@@ -160,7 +160,7 @@ export default {
         return this.value;
       },
       set(localValue) {
-       this.$emit("update:value", localValue);
+        this.$emit("update:value", localValue);
       },
     },
     isTitlePresent() {
@@ -263,9 +263,8 @@ export default {
     keyPress(event) {
       // invoked when a key is pressed
       // character key not accepted
-      event = event || window.event;
-      var charCode = (typeof event.which == "undefined") ? event.keyCode : event.which;
-      var charStr = String.fromCharCode(charCode);
+      let charCode = event.keyCode;
+      let charStr = String.fromCharCode(charCode);
       if (!charStr.match(/^[0-9]+$/)){
         event.preventDefault();
       }
--- a/src/components/UI/Text/InputNumber.vue
+++ b/src/components/UI/Text/InputNumber.vue
@@ -173,7 +173,13 @@ export default {
     },
     isValidationEnabled() {
       // whether input validation is on
-      return this.validation["enabled"] && this.validationMessage != "";
+      if(this.localValue.toString() == "0")
+      {
+        return this.validationMessage != "";
+      }
+      else{
+        return this.validation["enabled"] && this.validationMessage != "";
+      }
     },
     isValid() {
       // whether the input is valid
--- a/src/components/UI/Text/InputNumber.vue
+++ b/src/components/UI/Text/InputNumber.vue
@@ -53,7 +53,7 @@
         :class="[inputAreaClass, boxStyling]"
         :min="min"
         :max="max"
-        type="tel"
+        type="type"
         autocomplete="off"
         data-test="input"
         :disabled="isDisabled"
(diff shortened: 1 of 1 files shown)


src/components/UI/Text/InputNumber.vue
```diff
@@ -147,6 +147,7 @@ export default {
       default: false,
       type: Boolean,
     },
+
```
dalmia, line 150: this seems to be a change that happened by mistake?

src/components/UI/Text/InputNumber.vue
```diff
@@ -257,7 +258,7 @@ export default {
     },
     keyPress(event) {
       // invoked when a key is pressed
-      if (this.maxLength != null && this.localValue.length == this.maxLength) {
+      if (this.maxLength != null && this.localValue.toString().length == this.maxLength) {
```
dalmia, line 261: wouldn't it be better to change the localValue to string where it is computed in the first place instead of converting it to string every time?

Salonichoudhary07: yes done

Salonichoudhary07: returning `localvalue` as a string from get method

src/components/UI/Text/InputNumber.vue
```diff
@@ -152,7 +152,7 @@ export default {
     localValue: {
       // local copy of the value prop
       get() {
-        return this.value;
+        return this.value.toString();
```
deepansh96, line 155: @dalmia are we sure we should do this?

Reason - 
this is an InputNumber component. The value that the user types in the textbox gets saved as a `Number` data type which makes sense because this is a InputNumber component.

We need to convert it into string only when doing this check `this.localValue.length == this.maxLength`

If we go ahead with converting the `localValue` into string as what Saloni did in her latest commit, then we're forever converting this number into a string and that is what gets emitted into the parent.

Currently it should work fine because we're not using this component anywhere else other than Login page but this doesn't feel correct given this is an inputNumber component. The component should hold an internal data as Number not a string.

deepansh96 (commented):
There's one more bug which you can fix in this PR itself @Salonichoudhary07 

When I type mobile number starting 0 in the text box, the text "Mobile number is invalid" doesn't show up. It shows up in all other cases where the number I type starts with 1, 2, 3 ... 9

src/components/UI/Text/InputNumber.vue
deepansh96: @Salonichoudhary07 I talked to Aman regarding this. He agrees with what I said here. 
The first commit you made, where you did `this.localValue.toString().lenght == this.`, we should do that change.

Also one more point of feedback that Aman asked me to give you, is that, if someone has reviewed your PR and asked something, that doesn't imply that it is the right answer. You should be confident of the change you made and if you're not, play around with things enough so that you're confident. In this case for example, Aman asked a question "If doing that would be better?". So you should defend the changes you've made :p

Salonichoudhary07: Yes, I will do that.

src/components/UI/Text/InputNumber.vue
```diff
@@ -173,7 +173,13 @@ export default {
...
     isValidationEnabled() {
       // whether input validation is on
-      return this.validation["enabled"] && this.validationMessage != "";
+      if(this.localValue.toString() == "0")
+      {
+        return this.validationMessage != "";
+      }
+      else{
+        return this.validation["enabled"] && this.validationMessage != "";
+      }
```
dalmia, line 182: this will still lead to the issue that I can add as many trailing zeros as I want and the eventual phone number length goes way beyond 10 and it is still marked valid (as shown in the image - no error message)

Screenshot 2022-01-31 at 10 18 00 AM

src/components/UI/Text/InputNumber.vue
```diff
@@ -53,7 +53,7 @@
         :class="[inputAreaClass, boxStyling]"
         :min="min"
         :max="max"
-        type="number"
+        type="type"
```
deepansh96, line 56: Can you help me understand what is this supposed to do?
As far as I remember, There's no "type" type in an HTML `<input>` tag

deepansh96: See this - https://www.w3schools.com/html/html_form_input_types.asp

Salonichoudhary07: I hope the change answers the question

src/components/UI/Text/InputNumber.vue
```diff
@@ -257,7 +262,14 @@ export default {
     },
     keyPress(event) {
       // invoked when a key is pressed
-      if (this.maxLength != null && this.localValue.length == this.maxLength) {
+      // character key not accepted
+      event = event || window.event;
```
deepansh96, line 266: What is the purpose of `window.event`?

src/components/UI/Text/InputNumber.vue
```diff
@@ -257,7 +262,14 @@ export default {
...
+      var charCode = (typeof event.which == "undefined") ? event.keyCode : event.which;
```
deepansh96, line 267: You're using `event.which` here.
If you read the docs for this, they give a warning that this is a non standard and should not be used in production as it's result is highly variable. See this - https://developer.mozilla.org/en-US/docs/Web/API/UIEvent/which

Did you find this piece of code online somewhere? It's okay if you did, but make you understand what's happening.

src/components/UI/Text/InputNumber.vue
```diff
@@ -257,7 +262,14 @@ export default {
...
+      var charStr = String.fromCharCode(charCode);
```
deepansh96, line 268:
```suggestion
      let charCode = (typeof event.which == "undefined") ? event.keyCode : event.which;
      let charStr = String.fromCharCode(charCode);
```

We should be using `let` here instead of `var`
You can read up on which one to use when

src/components/UI/Text/InputNumber.vue
deepansh96, line 56: prop not bound

src/components/UI/Text/InputNumber.vue
```diff
@@ -106,6 +106,11 @@ export default {
       default: "",
       type: [String, Number],
     },
+    /** the type of the input in the input box */
+    type: { 
+      default: 'tel' ,
```
deepansh96, line 111: Why this space between `tel` and the comma?

src/components/UI/Text/InputNumber.vue
```diff
@@ -155,7 +160,7 @@ export default {
         return this.value;
       },
       set(localValue) {
-        this.$emit("update:value", localValue);
+       this.$emit("update:value", localValue);
```
deepansh96, line 163: Always perform a self review before giving a pr to someone else for review. 
All you have to do is go to the "files changed" tab and see your changes yourself. 
You'll figure out these small small issues on your own.
Like here, this removal of space is not necessary here.

deepansh96 (changes requested):

src/components/UI/Text/InputNumber.vue
Salonichoudhary07: okay

src/components/UI/Text/InputNumber.vue
Salonichoudhary07: let allows us to declare variables that are limited in scope to the block, statement, or expression on which it is used. 
var keyword is used to define a variable globally, or locally to an entire function regardless of block scope.
Will use let here.

src/components/UI/Text/InputNumber.vue
Salonichoudhary07: Yes, I did, I went through the code I did not knew about the warning.

src/components/UI/Text/InputNumber.vue
Salonichoudhary07: It returns the Event which is currently being handled by the site's code, not required here as we are already specifying the event.

src/components/UI/Text/InputNumber.vue
Salonichoudhary07: Removed it as we only need the keycode to compare the keypress.
Thank you for your feedback, will surely keep these things in mind in the future.

src/components/UI/Text/InputNumber.vue
```diff
@@ -263,9 +263,8 @@ export default {
     keyPress(event) {
       // invoked when a key is pressed
       // character key not accepted
-      event = event || window.event;
-      var charCode = (typeof event.which == "undefined") ? event.keyCode : event.which;
-      var charStr = String.fromCharCode(charCode);
+      let charCode = event.keyCode;
```
deepansh96, line 266: These things should always be checked. 
`keyCode` is deprecated
https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode

deepansh96 (changes requested):

src/components/UI/Text/InputNumber.vue
Salonichoudhary07: will use event.key will work the same

Salonichoudhary07: will use event.key will work the same

src/components/UI/Text/InputNumber.vue
```diff
@@ -257,7 +262,11 @@ export default {
     },
     keyPress(event) {
       // invoked when a key is pressed
-      if (this.maxLength != null && this.localValue.length == this.maxLength) {
+      // character key not accepted
+      if (!/^[0-9]+$/.test(event.key)){
+        event.preventDefault();
+      }
```
deepansh96, line 268: You can do this in the <input> tag itself. No need to write extra javascript for this
Take a look at this 
https://www.w3schools.com/tags/att_input_pattern.asp

deepansh96 (changes requested):

src/components/UI/Text/InputNumber.vue
Salonichoudhary07: yes I tried this earlier too it was not working for me

deepansh96: Okay no issues then

dalmia (approved):

PR #604, opened 2022-01-28
Adds support for OTP resend + fixes icon alignment

### LINKED ISSUE #600: Fix OTP resend flow + icon formatting
- currently, the user gets the option to instantly resend OTP, and once they have resent, they cannot ask for resending anymore.
- the icon has become a little off on the login page
Screenshot 2022-01-25 at 4 10 30 PM

What is desired:
- wait for 60 seconds before enabling resend button
- once resent, restart the 60 second timer for resending

Salonichoudhary07:
Fixes #600 

## Summary
 

- `place-self-center` added to `class` of `icon div` to align icon in the center.
- Added `startCountdown(seconds)` in the `method` in `Login.vue` , this function starts the 60 second timer first on click of `Request OTP button` then everytime when `Resend OTP button` is clicked. 
- `Counting` variable is used to know whether the timer is enabled or not.
- `Counter` variable gives the count of the timer.

## Test Plan

- [ ] ~Test Responsiveness~
   - [ ] Laptop (1200px)
   - [ ] Tablet (760px)
   - [ ] Phone (320px)
- [ ] ~Cross-Browser Testing~
   - [ ] Chrome
   - [ ] Firefox
- [X] Local Language Support
   - [X] Self-review
   - [X] Comments have been added appropriately
   - [X] Added relevant details like Labels/Projects/Milestones etc.
       - [ ] update `docs/ENV.md`
       - [ ] update DockerFile
- [ ] Testing
   - [X] Wrote tests
   - [X] Tested locally
   - [ ] Tested on staging
   - [ ] Tested on production
- [ ] ~Lighthouse Checks~


src/components/UI/Text/InputNumber.vue
```diff
@@ -29,7 +29,7 @@
       <!-- start icon -->
       <div
         v-if="isStartIconEnabled"
-        class="z-10 absolute font-xl text-blueGray-300 bg-transparent rounded text-base items-center w-5 inset-y-1/4 left-1.5"
```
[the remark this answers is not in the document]
dalmia: @deepansh96 since we are only using InputNumber in Login.vue, is it okay to make changes directly in the component itself?

src/pages/Login.vue
```diff
@@ -162,12 +162,12 @@
...
+        v-if="requestedOtp && !counting"
         data-test="resendOTP"
       ></icon-button>
-      <!-- text to show when OTP has been resent -->
-      <p v-if="resentOtp" class="text-center mt-2">
-        {{ $t("login.otp.resent") }}
+      <!-- text to show when OTP will be resent -->
+      <p v-if="counting" class="text-center mt-2">
+          {{$t("login.otp.timer")}} {{ counter }} {{$t("login.otp.secs")}}
       </p>
```
dalmia, line 171: you do not need to create another element for showing the resent text - you can simply update the text of the resend button + add the classes `opacity-50` and `cursor-pointer-not-allowed` to the resend button class if the timer is being shown!

Also, we decided on using brackets right? - `Resend OTP (60 secs)` instead of `Resend OTP in 60 seconds`

src/pages/Login.vue
```diff
@@ -162,12 +162,12 @@
         :buttonClass="resendOTPButtonClass"
         class="mt-2"
         :isDisabled="submitOTPIconConfig.enabled"
-        v-if="requestedOtp && !resentOtp"
+        v-if="requestedOtp && !counting"
```
dalmia, line 165: this will change accordingly as you'll need to always show this resend button but the styling and the text based on whether the counter is running or not.

src/pages/Login.vue
```diff
@@ -224,6 +224,8 @@ export default {
     return {
       phoneInput: "", // phone input text
       otpInput: "", // otp input text
+      counting: false, // whether counter is enabled for resend OTP
+      counter: "", // the count of the counter
```
dalmia, line 228: `counting` is dependent on `counter` right? So, it should be a computed property!

src/pages/Login.vue
```diff
@@ -224,6 +224,8 @@ export default {
     return {
       phoneInput: "", // phone input text
       otpInput: "", // otp input text
+      counting: false, // whether counter is enabled for resend OTP
```
dalmia, line 227: better variable name (+ update comment):
`isResendOTPEnabled`

src/pages/Login.vue
dalmia, line 228: better variable name (+ update comment):
`resendOTPTimer`

src/pages/Login.vue
```diff
@@ -395,16 +397,32 @@ export default {
...
+            this.counting= false;
+            clearInterval(interval);
+          }
+      }, 1000);
+    },
     requestOtp() {
       // requests OTP for the first time
       UserAPIService.requestOtp(this.formattedPhoneInput);
       this.requestedOtp = true;
+      this.counting = true;
```
dalmia, line 416: this will no longer need to be set when it is made a computed property

src/pages/Login.vue
```diff
@@ -395,16 +397,32 @@ export default {
...
+      this.startCountdown(60);
       this.invalidOtp = false;
     },
     resendOtp() {
       // resends OTP on user request
       UserAPIService.requestOtp(this.formattedPhoneInput);
       this.resentOtp = true;
+      this.counting = true;
```
dalmia, line 424: this will no longer need to be set when it is made a computed property

src/pages/Login.vue
```diff
@@ -395,16 +397,32 @@ export default {
       // whether the phone number entered by the user is valid
       return this.phoneInput.toString().match(/^([0]|\+91)?[6-9]\d{9}$/g) != null;
     },
+    startCountdown(seconds) {
```
dalmia, line 400: better name + since you are going to use 60 only, you can use it as the default value and so, you wouldn't have to pass `60` while calling the function.

```suggestion
    startResendOTPTimer(seconds = 60) {
```

src/pages/Login.vue
```diff
@@ -395,16 +397,32 @@ export default {
...
+            clearInterval(interval);
+          }
+      }, 1000);
+    },
     requestOtp() {
       // requests OTP for the first time
       UserAPIService.requestOtp(this.formattedPhoneInput);
       this.requestedOtp = true;
+      this.counting = true;
+      this.startCountdown(60);
```
dalmia, line 417: based on feedback above:
```suggestion
      this.startResendOTPTimer();
```

src/pages/Login.vue
```diff
@@ -395,16 +397,32 @@ export default {
...
       this.invalidOtp = false;
     },
     resendOtp() {
       // resends OTP on user request
       UserAPIService.requestOtp(this.formattedPhoneInput);
       this.resentOtp = true;
+      this.counting = true;
+      this.startCountdown(60);
```
dalmia, line 425: based on feedback above:
```suggestion
      this.startResendOTPTimer();
```

src/pages/Login.vue
```diff
@@ -395,16 +397,32 @@ export default {
       // whether the phone number entered by the user is valid
       return this.phoneInput.toString().match(/^([0]|\+91)?[6-9]\d{9}$/g) != null;
     },
+    startCountdown(seconds) {
+      // counts seconds before enabling resend OTP button
```
dalmia, line 401: we have not updated the docstrings in this file yet but refer to the function `addNewItem` in `Editor.vue` for how we are writing docstrings for functions.

src/pages/Login.vue
```diff
@@ -395,16 +397,32 @@ export default {
...
+    startCountdown(seconds) {
+      // counts seconds before enabling resend OTP button
+      this.counter = seconds;
+      const interval = setInterval(() => {
+          this.counter--;
+          if(this.counter == 0)
+          {
+            this.counting= false;
+            clearInterval(interval);
+          }
```
dalmia, line 409: once you create the computed property, won't have to set it manually.
```suggestion
          if (!this.counter) clearInterval(interval);
```

src/pages/Login.vue
```diff
@@ -395,16 +397,32 @@ export default {
       // whether the phone number entered by the user is valid
       return this.phoneInput.toString().match(/^([0]|\+91)?[6-9]\d{9}$/g) != null;
     },
+    startCountdown(seconds) {
+      // counts seconds before enabling resend OTP button
+      this.counter = seconds;
+      const interval = setInterval(() => {
+          this.counter--;
+          if(this.counter == 0)
```
dalmia, line 405: plus, have you set up your Vue linter and `pre-commit`? - issues of wrong spacing (between `if` and `(this.counter == 0)`) should be fixed before the commit is made itself so that we don't have to point it out.

tests/unit/pages/Login.spec.js
```diff
@@ -96,21 +96,38 @@ describe("Login.vue", () => {
         const requestOtp = jest
           .spyOn(UserAPIService, "requestOtp")
           .mockImplementation(() => jest.fn());
-
+        const startCountdown = jest.spyOn(Login.methods, "startCountdown");
```
dalmia, line 99: all this will have to be changed once your implementation changes.

dalmia (changes requested):

src/components/UI/Text/InputNumber.vue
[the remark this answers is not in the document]
Salonichoudhary07: I am just aligning the icon in the center I believe that will be necessary where ever this component is used.

src/pages/Login.vue
Salonichoudhary07: Yes I checked for pre-commit

src/pages/Login.vue
Salonichoudhary07: I did this because I have to set counting to false.

src/pages/Login.vue
Salonichoudhary07: No, I don't have vue linter.

src/pages/Login.vue
Salonichoudhary07: It is looking better without brackets so that's why I didn't do that.

tests/unit/pages/Login.spec.js
```diff
@@ -96,21 +96,40 @@ describe("Login.vue", () => {
...
-
+        const startResendOTPTimer = jest.spyOn(
+          Login.methods,
+          "startResendOTPTimer"
+        );
         mountWrapper();
+
         await setPhoneNumber();
         await requestOTP();
+        await wrapper.setData({ isResendOTPEnabled: false });
```
dalmia, line 107: this should be false by default no?

tests/unit/pages/Login.spec.js
```diff
@@ -96,21 +96,40 @@ describe("Login.vue", () => {
...
         expect(requestOtp).toHaveBeenCalled();
+        expect(startResendOTPTimer).toHaveBeenCalled();
+        expect(wrapper.vm.resendOTPTimer).toBe(60);
         expect(wrapper.vm.resentOtp).toBe(true);
         expect(wrapper.vm.invalidOtp).toBe(false);
       });
 
+      it("OTP timer", async () => {
+        //To test the Resend OTP Timer
+        //Faketimer is only needed for this test case to advance the time of interval
```
dalmia, line 123: you do not need to describe what generic functions are used for - you can delete this

tests/unit/pages/Login.spec.js
```diff
@@ -96,21 +96,40 @@ describe("Login.vue", () => {
...
 
         expect(requestOtp).toHaveBeenCalled();
+        expect(startResendOTPTimer).toHaveBeenCalled();
+        expect(wrapper.vm.resendOTPTimer).toBe(60);
         expect(wrapper.vm.resentOtp).toBe(true);
         expect(wrapper.vm.invalidOtp).toBe(false);
       });
 
+      it("OTP timer", async () => {
+        //To test the Resend OTP Timer
```
dalmia, line 122: the name of the function should easily clarify what it does so that you should not need to add more comments.

tests/unit/pages/Login.spec.js
```diff
@@ -96,21 +96,40 @@ describe("Login.vue", () => {
...
         await flushPromises();
 
         expect(requestOtp).toHaveBeenCalled();
+        expect(startResendOTPTimer).toHaveBeenCalled();
+        expect(wrapper.vm.resendOTPTimer).toBe(60);
         expect(wrapper.vm.resentOtp).toBe(true);
         expect(wrapper.vm.invalidOtp).toBe(false);
       });
 
+      it("OTP timer", async () => {
```
dalmia, line 121: better name for the test. See the pattern in the name of the other tests.

tests/unit/pages/Login.spec.js
```diff
@@ -96,21 +96,40 @@ describe("Login.vue", () => {
...
+        //To test the Resend OTP Timer
+        //Faketimer is only needed for this test case to advance the time of interval
+        jest.useFakeTimers();
+        wrapper.vm.startResendOTPTimer(1);
+        jest.advanceTimersByTime(1000);
+        expect(wrapper.vm.isResendOTPEnabled).toBe(false);
+        expect(wrapper.vm.resendOTPTimer).toBe(0);
+        //Rest of code should work in real time
```
dalmia, line 129: just simply 
```suggestion
        // reset timers
```

tests/unit/pages/Login.spec.js
```diff
@@ -96,21 +96,40 @@ describe("Login.vue", () => {
...
         expect(wrapper.vm.invalidOtp).toBe(false);
       });
 
+      it("OTP timer", async () => {
+        //To test the Resend OTP Timer
+        //Faketimer is only needed for this test case to advance the time of interval
+        jest.useFakeTimers();
+        wrapper.vm.startResendOTPTimer(1);
+        jest.advanceTimersByTime(1000);
+        expect(wrapper.vm.isResendOTPEnabled).toBe(false);
```
dalmia, line 127: once the timer is done, logically the button SHOULD BE enabled right?

tests/unit/pages/Login.spec.js
```diff
@@ -96,21 +96,40 @@ describe("Login.vue", () => {
...
+        expect(wrapper.vm.resendOTPTimer).toBe(60);
         expect(wrapper.vm.resentOtp).toBe(true);
         expect(wrapper.vm.invalidOtp).toBe(false);
       });
 
+      it("OTP timer", async () => {
+        //To test the Resend OTP Timer
+        //Faketimer is only needed for this test case to advance the time of interval
+        jest.useFakeTimers();
+        wrapper.vm.startResendOTPTimer(1);
```
dalmia, line 125: since your loop runs after every 1 second, best to use a value > 1 second.

tests/unit/pages/Login.spec.js
```diff
@@ -96,21 +96,40 @@ describe("Login.vue", () => {
...
+        jest.advanceTimersByTime(1000);
```
dalmia, line 126: make the value of seconds that you passed above a variable (say, `numSeconds`) and reuse it here (`numSeconds * 1000`)

dalmia (changes requested):

src/pages/Login.vue
```diff
@@ -278,6 +275,13 @@ export default {
         iconClass: "animate-spin h-4 object-scale-down text-white",
       };
     },
+    isResendOTPEnabled() {
+      //return whether resendOTPTimer is enabled for resend OTP
+      if (this.resendOTPTimer == 0 || this.resendOTPTimer == 60) {
```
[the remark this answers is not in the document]
Salonichoudhary07: Yes, I got confused.

src/pages/Login.vue
```diff
@@ -224,6 +220,7 @@ export default {
     return {
       phoneInput: "", // phone input text
       otpInput: "", // otp input text
+      resendOTPTimer: "", // the count of the timer
```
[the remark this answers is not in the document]
Salonichoudhary07: Yes

src/pages/Login.vue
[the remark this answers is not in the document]
Salonichoudhary07: I changed the true and false. Does it make sense now?

src/pages/Login.vue
```diff
@@ -395,16 +407,32 @@ export default {
       // whether the phone number entered by the user is valid
       return this.phoneInput.toString().match(/^([0]|\+91)?[6-9]\d{9}$/g) != null;
     },
+    startResendOTPTimer(seconds = 60) {
+      /**
+       *  counts seconds before enabling resend OTP button
+       *
+       * @param {Integer} seconds - how many seconds timer to be executed
```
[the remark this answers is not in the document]
Salonichoudhary07: Yes sorry

tests/unit/pages/Login.spec.js
Salonichoudhary07: It was to test the functioning of the timer.

tests/unit/pages/Login.spec.js
Salonichoudhary07: Yes, it is enabled, I had used true and false in the opposite sense, I have changed it now.

tests/unit/pages/Login.spec.js
Salonichoudhary07: image
Then there comes a testing error in google authentication.
And also we just need to loop it to see that after the timer is ending the button is enabling,1 works for that.

tests/unit/pages/Login.spec.js
Salonichoudhary07: Yes, this can be done.

tests/unit/pages/Login.spec.js
Salonichoudhary07: And to check enabling of button

src/pages/Login.vue
```diff
@@ -275,12 +273,14 @@ export default {
...
+    /**
+     * whether resendOTPTimer is enabled for resend OTP
+     */
     isResendOTPEnabled() {
-      //return whether resendOTPTimer is enabled for resend OTP
-      if (this.resendOTPTimer == 0 || this.resendOTPTimer == 60) {
-        return false;
+      if (this.resendOTPTimer == 0) {
+        return true;
       }
```
dalmia, line 282: single lines don't need a block
```suggestion
      if (this.resendOTPTimer == 0) return true;
```

src/pages/Login.vue
```diff
@@ -275,12 +273,14 @@ export default {
         iconClass: "animate-spin h-4 object-scale-down text-white",
       };
     },
+    /**
+     * whether resendOTPTimer is enabled for resend OTP
```
dalmia, line 277: this is the final use case of this right?
```suggestion
     * whether the button for resending OTP is enabled
```

src/pages/Login.vue
```diff
@@ -222,7 +221,6 @@ export default {
       otpInput: "", // otp input text
       resendOTPTimer: "", // the count of the timer
```
dalmia, line 222: the default value for this is still a string but is meant to old integer values.

tests/unit/pages/Login.spec.js
```diff
@@ -114,19 +112,19 @@ describe("Login.vue", () => {
...
-        wrapper.vm.startResendOTPTimer(1);
-        jest.advanceTimersByTime(1000);
-        expect(wrapper.vm.isResendOTPEnabled).toBe(false);
+        const numseconds = 2;
+        wrapper.vm.startResendOTPTimer(numseconds);
+        jest.advanceTimersByTime(numseconds * 1000);
+        expect(wrapper.vm.isResendOTPEnabled).toBe(true);
         expect(wrapper.vm.resendOTPTimer).toBe(0);
-        //Rest of code should work in real time
+        //reset timers
```
dalmia, line 127: please set up the linter so that such issues are caught before the commit itself.
```suggestion
        // reset timers
```

dalmia (changes requested):

dalmia:
the coverage for this function has been removed!

Screenshot 2022-02-08 at 2 38 59 AM

tests/unit/pages/Login.spec.js
```diff
@@ -107,10 +110,24 @@ describe("Login.vue", () => {
         await flushPromises();
 
         expect(requestOtp).toHaveBeenCalled();
-        expect(wrapper.vm.resentOtp).toBe(true);
+        expect(startResendOTPTimer).toHaveBeenCalled();
+        expect(wrapper.vm.resendOTPTimer).toBe(60);
+        expect(wrapper.vm.isResendOTPEnabled).toBe(false);
```
dalmia, line 115: you'll likely have to advance jest timer by 60 seconds so that the resend otp button becomes active so that the trigger at line 108 works - currently that trigger is not working.

dalmia:
Very minor bug.

steps to reproduce:
- enter a phone number
- request otp 
- before the timer gets over, delete some digits from the phone number
- request OTP again

what you will see:
- the timer is moving much more quickly than 1 second
- once the timer reaches 0, it starts becoming negative

https://user-images.githubusercontent.com/13185699/152873839-b915a643-2cae-482c-bce5-6abb5307db1b.mov

src/pages/Login.vue
Salonichoudhary07: yes

src/pages/Login.vue
Salonichoudhary07: sorry i thought i changed it

tests/unit/pages/Login.spec.js
Salonichoudhary07: if the trigger was not working then how come resendOTPTimer value was being set?
I didn't understand this.

src/pages/Login.vue
```diff
@@ -274,12 +274,10 @@ export default {
       };
     },
     /**
-     * whether resendOTPTimer is enabled for resend OTP
+     *  whether the button for resending OTP is enabled
```
dalmia, line 277: extra space here!

src/pages/Login.vue
```diff
@@ -274,12 +274,10 @@ export default {
...
     },
     /**
-     * whether resendOTPTimer is enabled for resend OTP
+     *  whether the button for resending OTP is enabled
      */
     isResendOTPEnabled() {
-      if (this.resendOTPTimer == 0) {
-        return true;
-      }
+      if (this.resendOTPTimer == 0) return true;
```
dalmia, line 280:
```suggestion
      if (!this.resendOTPTimer) return true;

```

src/pages/Login.vue
```diff
@@ -416,7 +414,7 @@ export default {
       this.resendOTPTimer = seconds;
       const interval = setInterval(() => {
         this.resendOTPTimer--;
-        if (!this.resendOTPTimer) clearInterval(interval);
+        if (!this.resendOTPTimer || !this.requestedOtp) clearInterval(interval);
```
dalmia, line 417: a better solution would be to call clearInterval as soon as the `requestedOtp` value changes to false (based on its watcher) because here there is still a possibility that one might see resend otp = -1 - very slim chance but possible

tests/unit/pages/Login.spec.js
```diff
@@ -95,36 +94,41 @@ describe("Login.vue", () => {
...
+        expect(resendOtp).toHaveBeenCalled();
         expect(startResendOTPTimer).toHaveBeenCalled();
         expect(wrapper.vm.resendOTPTimer).toBe(60);
         expect(wrapper.vm.isResendOTPEnabled).toBe(false);
         expect(wrapper.vm.invalidOtp).toBe(false);
       });
 
       it("enables ResendOTP button when timer ends", async () => {
-        //Faketimer is only needed for this test case to advance the time of interval
+        //Faketimer is needed for this test case to advance the time of interval
```
dalmia, line 124:
```suggestion
        // fake timer is needed for this test case to advance the time of interval
```

tests/unit/pages/Login.spec.js
```diff
@@ -95,36 +94,41 @@ describe("Login.vue", () => {
...
         const startResendOTPTimer = jest.spyOn(
           Login.methods,
           "startResendOTPTimer"
         );
         mountWrapper();
 
         await setPhoneNumber();
+        jest.useFakeTimers();
         await requestOTP();
+        await jest.advanceTimersByTime(60000);
```
dalmia, line 107: Great! 👍

dalmia (changes requested):

src/pages/Login.vue
Salonichoudhary07: Okay I have made changes to it now, it will clear the interval as soon as the phone number is invalid.

Salonichoudhary07: that is going to work in every case.

src/pages/Login.vue
```diff
@@ -412,22 +417,24 @@ export default {
...
-        if (!this.resendOTPTimer || !this.isPhoneValid()) clearInterval(interval);
+        if (!this.resendOTPTimer) clearInterval(this.interval);
       }, 1000);
     },
     requestOtp() {
       // requests OTP for the first time
       UserAPIService.requestOtp(this.formattedPhoneInput);
       this.requestedOtp = true;
       this.startResendOTPTimer();
+      this.otpInput= "";
```
dalmia, line 428: same space issue! Please try looking up for some tool that can automatically correct this or please be mindful while writing the code itself that you do not make this error. doesn't feel like a good use of our time to keep pointing this out :/

src/pages/Login.vue
```diff
@@ -412,22 +417,24 @@ export default {
...
       this.invalidOtp = false;
     },
     resendOtp() {
       // resends OTP on user request
       UserAPIService.requestOtp(this.formattedPhoneInput);
       this.startResendOTPTimer();
+      this.otpInput= "";
```
dalmia, line 435: same space issue

src/pages/Login.vue
```diff
@@ -214,12 +214,17 @@ export default {
...
+    {
+       if (!this.resendOTPTimer) clearInterval(this.interval);
+    },
   },
   data() {
     return {
       phoneInput: "", // phone input text
       otpInput: "", // otp input text
       resendOTPTimer: 0, // the count of the timer
+      interval: 0, // to reset the timer
```
dalmia, line 225: clearer name and description - please spend significant time thinking about variable names, function names and comments - whether they would be clear to someone who is reading your code for the first time!

```
otpTimerInterval: 0, // to reset the otp timer
```

src/pages/Login.vue
```diff
@@ -214,12 +214,17 @@ export default {
         this.requestedOtp = false;
       }
     },
+    requestedOtp()
+    {
+       if (!this.resendOTPTimer) clearInterval(this.interval);
```
dalmia, line 217: this should be run if the timer is not = 0 right?

src/pages/Login.vue
Salonichoudhary07: Yes, I don't know why it is not working will find something else.

src/pages/Login.vue
```diff
@@ -214,15 +214,15 @@ export default {
     },
     requestedOtp()
     {
-       if (!this.resendOTPTimer) clearInterval(this.interval);
+       if (!this.requestedOtp) clearInterval(this.otpTimerInterval);
```
dalmia, line 217: the condition is not correct here - please don't rush - think through slowly!

Salonichoudhary07: earlier you said to watch requestOtp when it turns false

Salonichoudhary07: when we change the input the `requestedOTP` value will be changed to false

Salonichoudhary07: the watcher will see and will not have to wait

src/pages/Login.vue
Salonichoudhary07: For timer we are seeing it below in startCountdown function

src/pages/Login.vue
Salonichoudhary07: a better solution would be to call clearInterval as soon as the requestedOtp value changes to false (based on its watcher) because here there is still a possibility that one might see resend otp = -1 - very slim chance but possible. I made changes according to this.

dalmia (approved):

PR #615, opened 2022-02-14
BUG: Execution stopped for SSO users because of a bug in the settings implementation

### LINKED ISSUE #614: BUG: Execution stopped for SSO users because of a bug in the settings implementation
In the settings current implementation, workspace settings would stay null if a user is not part of any organization. This never caused a problem because for a personal workspace, it's given that workspace settings would not exist and for a non personal workspace, workspace settings would exist.

But a user logged in through SSO lies somewhere in the middle. That user is not PART of any organization, so its organizations array will be empty as received from the backend. The issue comes when the getter `isPersonalWorkspace` is being used in different places. For an SSO logged in user, the `isPersonalWorkspace` is `false` but it doesn't contain any workspace settings.
This leads to a an exception being thrown at a lot of places.

deepansh96:
Fixes #614 

## Summary
- For a user logged in through SSO, we'll keep the `isPersonalWorkspace` as true. Doing this will avoid the code using `workspaceSettings` in different places because `workspaceSettings` for an SSO Logged in user will be null as the user is not ACTUALLY a member of that organization

## Test Plan

- [ ] ~Test Responsiveness~
   - [ ] Laptop (1200px)
   - [ ] Tablet (760px)
   - [ ] Phone (320px)
- [x] Cross-Browser Testing
   - [ ] Chrome
   - [ ] Firefox
   - [x] Self-review
       - [ ] update `docs/ENV.md`
       - [ ] update DockerFile
- [ ] Testing
   - [x] Wrote tests
   - [x] Tested locally
   - [ ] Tested on staging
   - [ ] Tested on production
- [ ] ~Lighthouse Checks~


src/store/modules/auth.js
```diff
@@ -65,6 +68,11 @@ const getters = {
   /** settings for the active workspace */
   activeWorkspaceSettings: (state, getters) => {
     if (getters.isPersonalWorkspace) return null;
+    if (
+      state.workspaceSettings == null ||
```
dalmia, line 72: merge this with the if above

dalmia (approved):

PR #623, opened 2022-02-15
Add QuillJS editor to textarea

Manyakalra3:
## Summary
-Added bold,italic,underline functionality in the `Textarea`  component using quilljs
-Replaced textarea with div in `textarea` component.

## TODO

- [x]  use emits and props for two way binding instead of v-model(to show data in preview )
- [x] adjust css of textarea.
- [x] add placeholder to div
- [x] adjust the scroll according to height of div
- [x] Image icon should attach itself to top whenever textarea size changes.
- [x] disable answer textarea when not in use
- [x] add editor to input options too.
- [x] adjust editor padding.
- [x] change B,I,U buttons colours to orange and a background colour to them.
- [x] Fix the player part so that it shows only formatted text.
- [x] Check formatted text for player and item modal.
- [x] test the disabled input part ,as input is now converted to div.
- [x] fix the delete option in inputText

### Code todos
- [x] adjusts spaces in code and add comment above methods and any included code.
- [x] remove contenteditable from `textareadiv`.
- [x] change `editor` to `quillEditor`
- [x] Add comment saying that importing css file is necessary and icons wont appear properly without it.
- [ ] add tailwind class in a variable
- [x] change `componentkey` to `rerenderkey`,add comments too.
- [x] change forceUpdate to `Rerender`.
- [x] change componentley from 0 to 1 and vice versa instead of increasing the key,

## Test Plan

- [x] Test Responsiveness
   - [x] Laptop (1200px)
   - [x] Tablet (760px)
   - [x] Phone (320px)
- [ ] Cross-Browser Testing
   - [x] Chrome
   - [ ] Firefox
   - [x] Self-review
       - [ ] update `docs/ENV.md`
       - [ ] update DockerFile
- [ ] Testing
   - [ ] Wrote tests
   - [x] Tested locally
   - [ ] Tested on staging
   - [ ] Tested on production
- [ ] ~Lighthouse Checks~


deepansh96:
- Why isn't the editor box taking the full horizontal space? Why is there some gap on the left side?

image
<hr>
- The icons are blue in color, whereas in the mockup, the were that orangish color. Also, In the mockup, only the icon is highlighting but don't you think it would be better if when selecting a mode, like Bold, the B icon gets orange colored as well as it gets a background, to show that it's still active

Screenshot 2022-02-17 at 4 31 35 PM
<hr>
- Fixing this is still left right? On the preview, we want to see the _formatted_ version. Not the HTML version

Screenshot 2022-02-17 at 4 33 12 PM
<hr>
- While sending the data through the API call, `<p>` tag in the very beginning and `</p>` tag at the very end are being appended. We don't need these tags in our final database right? This does not add any information. The bold, em, strong, such tags add information, but `<p>` tag doesn't add any information

image
<hr>
- I'm seeing the text formatter textarea inside the preview. This shouldn't happen right?
 
image

<hr>
- Just a reminder, implementing formatting on ItemModal component (the left preview panel in editor) will solve it for the player as well, because we use the same component

image
<hr>

src/components/UI/Text/Textarea.vue
```diff
@@ -36,26 +36,30 @@
...
         data-test="input"
-      />
+      ></div>
     </div>
   </div>
 </template>
 
 <script>
+import Quill from "quill";
+import "quill/dist/quill.snow.css";
```
deepansh96, line 61: Add comment saying that importing css file is necessary and icons wont appear properly without it

src/components/UI/Text/Textarea.vue
```diff
@@ -36,26 +36,30 @@
...
         <inline-svg :src="startIconObj"></inline-svg>
       </div>
-
       <!-- input text area -->
-      <textarea
-        class="p-2 border placeholder-blueGray-300 text-blueGray-600 bg-white disabled:bg-gray-200 rounded text-md border-blueGray-300 focus:outline-none focus:ring focus:border-transparent focus:shadow-outline w-full border-gray-200 disabled:cursor-not-allowed"
+
+      <div
+        contenteditable="true"
+        class="pt-7 h-30 textbox pl-0 border placeholder-blueGray-300 text-blueGray-600 bg-white disabled:bg-gray-200 rounded text-md border-blueGray-300 focus:outline-none focus:ring focus:border-transparent focus:shadow-outline w-full border-gray-200 disabled:cursor-not-allowed"
```
deepansh96, line 43: This has become a long enough class :p
I think you should put it in a variable and then link the variable here.
For reference, see the rest of code where we've added tailwind classes as variables and used in HTML

src/components/UI/Text/Textarea.vue
```diff
@@ -36,26 +36,30 @@
...
         :class="[inputAreaClass, boxStyling]"
         :disabled="isDisabled"
+        ref="editor"
```
deepansh96, line 46: Is `textarea` a better name?
or maybe `inputArea` or `inputTextArea`

I think `editor` is not the most accurate

Manyakalra3:
> * Why isn't the editor box taking the full horizontal space? Why is there some gap on the left side?
> 
> image
> 
> - The icons are blue in color, whereas in the mockup, the were that orangish color. Also, In the mockup, only the icon is highlighting but don't you think it would be better if when selecting a mode, like Bold, the B icon gets orange colored as well as it gets a background, to show that it's still active
> Screenshot 2022-02-17 at 4 31 35 PM
> 
> - Fixing this is still left right? On the preview, we want to see the _formatted_ version. Not the HTML version
> Screenshot 2022-02-17 at 4 33 12 PM
> 
> - While sending the data through the API call, `
> ` tag in the very beginning and `
> 
> ` tag at the very end are being appended. We don't need these tags in our final database right? This does not add any information. The bold, em, strong, such tags add information, but `
> ` tag doesn't add any information
> 
> image
> 
> - I'm seeing the text formatter textarea inside the preview. This shouldn't happen right?
> image
> 
> - Just a reminder, implementing formatting on ItemModal component (the left preview panel in editor) will solve it for the player as well, because we use the same component
> image

- for the left space I will adjust the padding .
- Yess,this sounds good,will discuss background colour with priyansha and will add that too.
- Yess,fixing is still left, I will use HTML parser on the frontend for that.
- By starting and ending tags, you mean `<p><br></p>` right>?
- Answer textarea willl be disabled,will make the changes.

src/components/UI/Text/Textarea.vue
```diff
@@ -114,6 +118,12 @@ export default {
       type: Number,
     },
   },
+  data() {
+    return {
+      editor: null,
```
deepansh96, line 123: this is an instance of the quilljs editor right?
maybe rename it to `quillInstance` or `quillEditor`?

deepansh96: Also add comments!

src/components/UI/Text/Textarea.vue
```diff
@@ -194,10 +204,12 @@ export default {
     },
   },
   methods: {
-    inputChange(event) {
+    update() {
+      this.$emit("update:value", this.editor.getText() ? this.editor.root.innerHTML : "");
```
deepansh96, line 208: Is there any case where `this.editor.getText()` will return a false boolean? I mean, is there any case where this if/else condition will return ""?

src/components/UI/Text/Textarea.vue
```diff
@@ -194,10 +204,12 @@ export default {
     },
   },
   methods: {
-    inputChange(event) {
+    update() {
```
deepansh96, line 207: What is being updated here?
You should name the method such that it's clear what action is it doing

deepansh96: Also add one-liner docstring above the method name

src/components/UI/Text/Textarea.vue
```diff
@@ -194,10 +204,12 @@ export default {
...
+      this.$emit("update:value", this.editor.getText() ? this.editor.root.innerHTML : "");
+    },
+    inputChange() {
       // invoked on input change
       this.$emit("input", this.value);
```
deepansh96, line 212: This is not being used right now right? I guess this is in your todo list

src/components/UI/Text/Textarea.vue
```diff
@@ -220,5 +233,32 @@ export default {
     },
   },
   emits: ["input", "keypress", "keydown", "update:value", "start-icon-selected"],
+  // quilljs editor
```
deepansh96, line 236: This comment is above the `mounted`.
Currently, in `mounted`, the only thing happening is setting up of quill editor.
But other things can also happen, not related to quill editor.
So keeping a comment `quilljs editor` above `mounted` is not correct as `mounted` is not something that we have defined. It's already there. We're just doing a bunch of stuff when component is mounted.

You should move this comment inside, just above the code where you're dealing with Quill editor. And write the comment in a little more detail.

src/components/UI/Text/Textarea.vue
```diff
@@ -220,5 +233,32 @@ export default {
...
+  mounted() {
+    this.editor = new Quill(this.$refs.editor, {
+      modules: {
+        toolbar: [["bold", "italic", "underline"]],
+      },
+      theme: "snow",
+      formats: ["bold", "underline", "italic"],
+    });
+
+    this.editor.root.innerHTML = this.value;
```
deepansh96, line 246: If you're using value here then I guess you don't the variable `localValue` anymore right?

src/components/UI/Text/Textarea.vue
```diff
@@ -220,5 +233,32 @@ export default {
...
+  position: absolute;
+  width: 100%;
+  background-color: #f2e8df;
+  border: 1px solid transparent;
+  padding: 0px;
+}
+.ql-editor {
+  overflow-y: auto;
+}
+</style>
```
deepansh96, line 264: is this still needed? Because I didn't see these classes being used anywhere

src/pages/Editor.vue
```diff
@@ -368,6 +368,7 @@
             </div>
             <!--- item editor  -->
             <item-editor
+              :key="componentKey"
```
deepansh96, line 371: this key is specific to the purpose of re-rendering right?
Better to keep consistency and do what we've done in rest of the codebase, named it `reRenderKey`

src/pages/Editor.vue
```diff
@@ -597,6 +598,7 @@ export default {
     const confettiHandler = confetti.create(confettiCanvas, { resize: true });
 
     return {
+      componentKey: 0,
```
deepansh96, line 601: add one liner comment

src/pages/Editor.vue
```diff
@@ -1386,6 +1388,9 @@ export default {
     },
   },
   methods: {
+    forceUpdate() {
```
deepansh96, line 1391: just `reRender() {` is fine no?

src/pages/Editor.vue
```diff
@@ -1386,6 +1388,9 @@ export default {
...
+      this.componentKey++;
```
deepansh96, line 1392: Not really needed to keep increasing the key. Just changing it from 0 to 1 or 1 to 0 is fine.

deepansh96 (changes requested):

src/components/UI/Text/Textarea.vue
```diff
@@ -25,7 +25,10 @@
       </div>
     </div>
 
-    <div class="flex relative mt-1">
+    <div
+    class=" flex relative mt-1"
+
+    >
```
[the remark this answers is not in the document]
Manyakalra3: I used ESlint as well as precmmit too!But they were not removed

src/components/UI/Text/Textarea.vue
```diff
@@ -36,26 +39,30 @@
       >
         <inline-svg :src="startIconObj"></inline-svg>
       </div>
-
       <!-- input text area -->
-      <textarea
-        class="p-2 border placeholder-blueGray-300 text-blueGray-600 bg-white disabled:bg-gray-200 rounded text-md border-blueGray-300 focus:outline-none focus:ring focus:border-transparent focus:shadow-outline w-full border-gray-200 disabled:cursor-not-allowed"
+
+      <div
+        contenteditable="true"
```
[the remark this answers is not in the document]
Manyakalra3: This was added to make div contenteditable. But yes since ql-editor is the child component and is editable ,this was not necessary.I will remove this.

src/components/UI/Text/Textarea.vue
```diff
@@ -36,26 +39,30 @@
...
+        class="pt-7 h-30 textbox pl-0 border placeholder-blueGray-300 text-blueGray-600 bg-white disabled:bg-gray-200 rounded text-md border-blueGray-300 focus:outline-none focus:ring focus:border-transparent focus:shadow-outline w-full border-gray-200 disabled:cursor-not-allowed"
         :class="[inputAreaClass, boxStyling]"
         :disabled="isDisabled"
+       ref="editor"
```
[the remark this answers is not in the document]
Manyakalra3: okayyy

src/components/UI/Text/Textarea.vue
Manyakalra3: sounds good! will make these changes

src/components/UI/Text/Textarea.vue
Manyakalra3: inputArea sounds good. will change this.

src/components/UI/Text/Textarea.vue
Manyakalra3: Instead of textarea or inputarea, I'll rename it as quillEditor then

src/components/UI/Text/Textarea.vue
Manyakalra3: okay

src/components/UI/Text/Textarea.vue
Manyakalra3: I cannot think of any case right now.But  I kept this to avoid edge cases if there are any,should I change it then?

src/components/UI/Text/Textarea.vue
Manyakalra3: Text is being updated here. I will change it to UpdateChange.
will add comments too!

src/components/UI/Text/Textarea.vue
Manyakalra3: This inputchange method is in use but the emit method inside it is not in need ,I will remove that.

src/components/UI/Text/Textarea.vue
Manyakalra3: No, I don't require that. Also, I was about to question why `localValue` was included in the code. Should I remove its instances from the code because we can do the functions with only the `value` variable?

src/components/UI/Text/Textarea.vue
Manyakalra3: Yes, it is required. This css is used by the buttons (bol, itlic, and underline). Except that the `ql-editor` css has yet to be applied.

src/pages/Editor.vue
Manyakalra3: I chose this name because there was already a key with the name `reRenderKey`, but yess,it should still work. I'll have another look at the code.

src/pages/Editor.vue
Manyakalra3: okay

src/pages/Editor.vue
Manyakalra3: reRender function name? yess,will change that

src/pages/Editor.vue
Manyakalra3: okayy,will do that.

PR #629, opened 2022-02-17
Prevents viewer from using the video seekbar to skip questions when skip disabled

### LINKED ISSUE #628: Viewers can use the video seekbar to skip questions when skip disabled

dalmia:
Fixes #628 

## Summary
- maintaining a variable for the last answered interaction index
- when the video is seeked, if the skip is disabled and there are interactions pending and the timestamp that is being seeked to is greater than the timestamp of the first unanswered interaction index, the viewer is shown that interaction
- when the page is reloaded, if the user was at a timestamp > timestamp of the first unanswered interaction index, we show that interaction.
- if a user clicks on play when on an unanswered question, they are not allowed to move ahead if skip is disabled
 
## Test Plan

- [ ] ~Test Responsiveness~
   - [ ] Laptop (1200px)
   - [ ] Tablet (760px)
   - [ ] Phone (320px)
- [x] Cross-Browser Testing
   - [ ] Chrome
   - [ ] Firefox
- [x] Local Language Support
   - [ ] Self-review
       - [ ] update `docs/ENV.md`
       - [ ] update DockerFile
- [ ] Testing
   - [ ] Wrote tests
   - [ ] Tested locally
   - [ ] Tested on staging
   - [ ] Tested on production
- [ ] ~Lighthouse Checks~

### DIFF
--- a/src/pages/Embeds/Plio.vue
+++ b/src/pages/Embeds/Plio.vue
@@ -930,12 +930,16 @@ export default {
       return linkValidation["ID"];
     },
     playerPlayed() {
+      console.log(this.player.playing);
+      console.log(this.isSkipEnabled);
+      console.log(this.isAnyItemActive);
       // invoked when the play button of the player is clicked
       if (this.isScorecardShown || (!this.isSkipEnabled && this.isAnyItemActive)) {
         /**
          * prevents the video from playing while the
          * scorecard is being shown or when an item is active with skip disabled
          */
+        console.log("yolo");
         this.pausePlayer();
         return;
       }
--- a/src/pages/Embeds/Plio.vue
+++ b/src/pages/Embeds/Plio.vue
@@ -146,6 +146,7 @@ export default {
           global: false,
         },
 
+        hideControls: false,
         clickToPlay: false,
 
         invertTime: false,
@@ -341,6 +342,10 @@ export default {
   computed: {
     ...mapGetters("auth", ["isAuthenticated"]),
     ...mapState("generic", ["windowInnerWidth", "windowInnerHeight"]),
+    currentItem() {
+      if (!this.isAnyItemActive) return null;
+      return this.items[this.currentItemIndex];
+    },
     numItems() {
       return this.items.length;
     },
@@ -454,7 +459,7 @@ export default {
      * eg - question, note etc
      */
     currentItemType() {
-      return this.items[this.currentItemIndex].type;
+      return this.currentItem.type;
     },
     /**
      * config for the title of the maximise button
@@ -681,24 +686,43 @@ export default {
         const firstUnansweredInteraction = this.items[
           this.lastAnsweredInteractionIndex + 1
         ];
-        if (timestamp > firstUnansweredInteraction.time)
+        if (timestamp >= firstUnansweredInteraction.time)
           return firstUnansweredInteraction;
       }
     },
-    videoSeeked() {
+    moveToFirstUnansweredItemTimestampOrPass() {
+      let playerProgress = document.getElementsByClassName("plyr__progress__buffer")[0];
+      console.log(playerProgress.getAttribute("value"));
+      let timeToInspect = this.player.currentTime;
+
+      if (
+        this.isAnyItemActive &&
+        this.player.currentTime < this.currentItem.time &&
+        this.player.currentTime >= this.currentItem.time - POP_UP_CHECKING_FREQUENCY
+      )
+        timeToInspect += POP_UP_CHECKING_FREQUENCY;
+
       const firstUnansweredInteraction = this.checkMovingToTimestampAllowed(
-        this.player.currentTime
+        timeToInspect
       );
       if (firstUnansweredInteraction != null) {
         this.setPlayerTime(firstUnansweredInteraction.time - POP_UP_CHECKING_FREQUENCY);
         setTimeout(() => {
           this.toast.error(this.$t("toast.player.cannot_skip_interaction"), {
-            id: "internetLostToast",
+            id: "cannotSkipInteraction",
             position: "bottom-center",
           });
         }, 500);
+        return true;
+      }
+      return false;
+    },
+    videoSeeked() {
+      if (this.moveToFirstUnansweredItemTimestampOrPass()) {
+        this.isModalMinimized = false;
         return;
       }
+
       // invoked when a seek operation ends
       this.createEvent("video_seeked", { currentTime: this.player.currentTime });
     },
@@ -812,6 +836,7 @@ export default {
     closeItemModal() {
       // invoked when the modal is to be closed
       this.currentItemIndex = null;
+      this.toast.dismiss("cannotSkipInteraction");
       this.playPlayer();
     },
     playPlayer() {
@@ -930,16 +955,17 @@ export default {
       return linkValidation["ID"];
     },
     playerPlayed() {
-      console.log(this.player.playing);
-      console.log(this.isSkipEnabled);
-      console.log(this.isAnyItemActive);
+      if (this.moveToFirstUnansweredItemTimestampOrPass()) {
+        this.pausePlayer();
+        return;
+      }
+
       // invoked when the play button of the player is clicked
-      if (this.isScorecardShown || (!this.isSkipEnabled && this.isAnyItemActive)) {
+      if (this.isScorecardShown) {
         /**
          * prevents the video from playing while the
-         * scorecard is being shown or when an item is active with skip disabled
+         * scorecard is being shown
          */
-        console.log("yolo");
         this.pausePlayer();
         return;
       }
@@ -1110,6 +1136,8 @@ export default {
         this.currentItemIndex = itemIndex;
         this.markItemSelected();
         this.createEvent("item_opened", { itemIndex: this.currentItemIndex });
+      } else {
+        this.closeItemModal();
       }
     },
     /**
--- a/src/pages/Embeds/Plio.vue
+++ b/src/pages/Embeds/Plio.vue
@@ -343,12 +343,7 @@ export default {
     ...mapGetters('auth', ['isAuthenticated']),
     ...mapState('generic', ['windowInnerWidth', 'windowInnerHeight']),
     firstUnansweredItem() {
-      if (
-        this.skipEnabled ||
-        this.lastAnsweredItemIndex == -1 ||
-        this.lastAnsweredItemIndex == this.numItems - 1
-      )
-        return null
+      if (this.skipEnabled || this.lastAnsweredItemIndex == this.numItems - 1) return null
       return this.items[this.lastAnsweredItemIndex + 1]
     },
     currentItem() {
@@ -919,15 +914,17 @@ export default {
           if (this.isItemMCQ(itemIndex)) {
             itemResponse.answer = parseInt(itemResponse.answer)
           }
-          if (
-            !this.isItemResponseEmpty(itemIndex, itemResponse.answer) &&
-            !this.isSkipEnabled
-          ) {
-            this.lastAnsweredItemIndex = itemIndex
-          }
           this.itemResponses.push(itemResponse)
         })
         if (!this.isSkipEnabled) {
+          // set the last answered item index
+          for (let [itemIndex, itemResponse] of this.itemResponses.entries()) {
+            if (!this.isItemResponseEmpty(itemIndex, itemResponse.answer))
+              this.lastAnsweredItemIndex = itemIndex
+            // https://github.com/avantifellows/plio-frontend/pull/629#issuecomment-1053819610
+            else break
+          }
+
           // check if any item before the current timestamp is unanswered
           // if there is, then update the timestamp to just before the
           // unanswered item
--- a/src/pages/Embeds/Plio.vue
+++ b/src/pages/Embeds/Plio.vue
@@ -477,7 +477,7 @@ export default {
      * whether there are any items
      */
     hasAnyItems() {
-      return this.items.length != 0;
+      return this.numItems != 0;
     },
     /**
      * whether any item is currently active
@@ -578,6 +578,7 @@ export default {
      * Show the scorecard on top of the player
      */
     popupScorecard() {
+      if (this.checkMovingToTimestampAllowed(this.player.duration) != null) return;
       if (!this.isScorecardShown) {
         this.isScorecardShown = true;
         var scorecardModal = document.getElementById("scorecardmodal");
@@ -773,7 +774,7 @@ export default {
           this.itemDetails = plioDetails.itemDetails || [];
           // setting numSkipped to number of items. This value will keep reducing
           // as numCorrect and numWrong are calculated
-          this.numSkipped = this.items.length;
+          this.numSkipped = this.numItems;
           this.plioDBId = plioDetails.plioDBId;
           this.videoId = this.getVideoIDfromURL(plioDetails.videoURL);
           this.plioTitle = plioDetails.plioTitle;
@@ -961,7 +962,7 @@ export default {
       this.$mixpanel.track("Visit Player", {
         "Plio UUID": this.plioId,
         "Plio Video Length": this.player.duration || 0,
-        "Plio Num Items": this.items.length || 0,
+        "Plio Num Items": this.numItems || 0,
       });
       // sets various properties based on the screen size
       this.setScreenProperties();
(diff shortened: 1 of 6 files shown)


deepansh96:
- There are two places where the tooltip comes on top of the active div in touch/mobile. One is when I click the `Preview` button. Once is when I click `Publish` button. Can you fix this in this PR itself? Should be a small change

- I had this weird bug on mobile phone where in preview mode and in player as well, the questions were not getting updated to green color after answering. I guess this is not because of this PR, but due to this, I was being redirected to the same question again and again because the code thought that no question has been answered.

https://user-images.githubusercontent.com/20169829/155518919-7e02378c-3ab6-4dc9-8000-0b44eae6321d.mp4


- What should we do for cases where, let's say someone has answered a plio where skip was enabled. They answered some questions and skipped some questions. Now, the creator changed the setting to "skip enabled" as false. Now when the user returns to the plio, the code will drop the user at the last unanswered question, even if there are some other unanswered questions in between. The code is written with the assumption that all questions will be answered in order. So I'm not sure what should the code ideally do in this case.


- Also, a separate bug, not for this PR. In subjective question textbox, if I keep typing something without any space, the submit button stays hidden idk why. Only if I press spacebar or close the mobile keyboard, the submit button gets active

https://user-images.githubusercontent.com/20169829/155519290-f9ff0e46-0996-4969-b94a-143c84a8ba93.mp4

- bug, not always replicable. But replicable on mobile phones. Sometimes, even if the coloring of the markers is correct, it doesn't allow me to go back into the green questions waala area whereas it should allow me. And sometimes it starts working fine.

https://user-images.githubusercontent.com/20169829/155519867-b279d80d-5a64-43aa-b08d-a7451b051621.mp4

src/pages/Embeds/Plio.vue
```diff
@@ -659,7 +686,47 @@ export default {
...
+        timeToInspect
+      );
+      if (firstUnansweredInteraction != null) {
+        this.setPlayerTime(firstUnansweredInteraction.time - POP_UP_CHECKING_FREQUENCY);
+        setTimeout(() => {
+          this.toast.error(this.$t("toast.player.cannot_skip_interaction"), {
+            id: "cannotSkipInteraction",
+            position: "bottom-center",
+          });
+        }, 500);
```
deepansh96, line 719: The settimeout here means that after the player time is set above, 500ms after it, the toast would pop up?
This works fine mostly but on a slow phone (shivani's office phone) what happens is that the toast pops up before.

When you do `this.setPlayerTime(`, plyr tries to buffer the video from that point. Until some amount has been buffered, the question modal that should pop up, doesn't pops up. 

Wouldn't it be a better solution to show this toast when an item has popped up instead of relying on 500ms?

src/pages/Embeds/Plio.vue
```diff
@@ -836,8 +907,26 @@ export default {
           if (this.isItemMCQ(itemIndex)) {
             itemResponse.answer = parseInt(itemResponse.answer);
           }
+          if (
+            !this.isItemResponseEmpty(itemIndex, itemResponse.answer) &&
+            !this.isSkipEnabled
+          ) {
+            this.lastAnsweredInteractionIndex = itemIndex;
+          }
```
deepansh96, line 915: This piece will cause some confusion in one of the edge cases that I've mentioned in the list of comments I've left on the PR. 
This assumes that no unanswered question can exist before an answered question?
Because in the case where there are 4 unanswered questions and then 1 answered question, it will make the last question as `lastAnsweredItemIndex` and the code will never force the user to answer the other questions.

I don't know what the solution should be.
Either we disallow people to change this setting after they publish as it can cause conflict in the responses.
Or we warn them that some older users will not be forced to answer.
Or we update this code to actually force the user to answer all the left out questions in this specific case

src/pages/Embeds/Plio.vue
```diff
@@ -836,8 +907,26 @@ export default {
...
+            this.currentTimestamp
+          );
+          // check if any item before the current timestamp is unanswered
+          // if there is, then update the timestamp to just before the
+          // unanswered item
+          if (firstUnansweredInteraction != null) {
+            this.currentTimestamp =
+              firstUnansweredInteraction.time - POP_UP_CHECKING_FREQUENCY;
+          }
+        }
```
deepansh96, line 929: This logic is what is needed to be checked everytime the video is seeked
Currently you're doing it when creating session, but it should also be run whenever it is seeked. This will solve the bug that I experienced on mobile phones (added video clip in the comments on the PR)

src/pages/Embeds/Plio.vue
```diff
@@ -1043,23 +1137,26 @@ export default {
...
 
-      this.checkForItemPopup(timestamp);
+      const itemIndex = this.checkForItemPopup(timestamp);
+      if (itemIndex != null) {
+        this.currentItemIndex = itemIndex;
+        this.markItemSelected();
+        this.createEvent("item_opened", { itemIndex: this.currentItemIndex });
+      } else {
+        this.closeItemModal();
+      }
```
deepansh96, line 1147: doubt - why was this added?

src/pages/Embeds/Plio.vue
```diff
@@ -1043,23 +1137,26 @@ export default {
...
-      this.currentItemIndex = ItemFunctionalService.checkItemPopup(
+      return ItemFunctionalService.checkItemPopup(
         timestamp,
         this.itemTimestamps,
         POP_UP_PRECISION_TIME
       );
-      if (this.currentItemIndex != null) {
-        this.markItemSelected();
-        this.createEvent("item_opened", { itemIndex: this.currentItemIndex });
-      }
```
deepansh96, line 1062: doubt - why was this removed?

deepansh96 (changes requested):

src/pages/Embeds/Plio.vue
```diff
@@ -202,6 +205,7 @@ export default {
       isAspectRatioChecked: false, // whether the check for aspect ratio has been done
       watchingEventDBId: null, // the DB id of the latest 'watching' event for a given session
       plioSettings: null, // stores this plio's settings
+      lastAnsweredInteractionIndex: -1, // index of the interaction that was last answered
```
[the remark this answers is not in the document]
dalmia: haan sorry makes sense! 👍

src/pages/Embeds/Plio.vue
```diff
@@ -511,7 +532,12 @@ export default {
   },
   methods: {
     ...mapActions("auth", ["setAccessToken", "setActiveWorkspace"]),
-
+    isItemResponseEmpty(itemIndex, answer) {
+      if (answer == null) return true;
+      if (this.itemDetails[itemIndex].type == "mcq") return isNaN(answer);
+      if (this.itemDetails[itemIndex].type == "checkbox") return answer.length == 0;
+      return false;
+    },
```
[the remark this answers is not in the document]
dalmia: uska is already covered in `answer == null` - warna it is non-empty - this is how it is implemented everywhere else as well jahan bhi we are checking if answer has been submitted or not.

src/pages/Embeds/Plio.vue
```diff
@@ -562,6 +588,7 @@ export default {
      * Show the scorecard on top of the player
      */
     popupScorecard() {
+      if (this.checkMovingToTimestampAllowed(this.player.duration) != null) return;
```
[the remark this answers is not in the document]
dalmia: haan naam I will change (good suggestion 👍)

boolean nahi hain kyunki it returns an object corresponding to the last answered item if there is any + if skip is disabled!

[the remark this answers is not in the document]
dalmia: without this, if the user drags the seekbar to the video ka end, toh before our check can run, the scorecard is popped up beyond for the scorecard pop-up there is no check to be run - plyr automatically triggers it as we reach the video end.

src/pages/Embeds/Plio.vue
```diff
@@ -659,7 +686,47 @@ export default {
         if (maximizeButton != undefined) this.mountOnFullscreenPlyr(maximizeButton);
       });
     },
+    checkMovingToTimestampAllowed(timestamp) {
+      if (!this.isSkipEnabled && this.lastAnsweredInteractionIndex < this.numItems - 1) {
```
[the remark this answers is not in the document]
dalmia: since ek hi jagah use ho raha - is it okay if I add a comment here instead of adding another variable to the global scope?

src/pages/Embeds/Plio.vue
dalmia: just for cleanup - I was earlier trying to use `checkForItemPopup` and it was doing more things than just checking for item popup - so, was hard to reuse - abhi although I didn't end up using `checkForItemPopup` as I was originally trying to, this cleanup still made sense, no?

src/pages/Embeds/Plio.vue
dalmia: explained above.

src/pages/Embeds/Plio.vue
```diff
@@ -659,7 +686,47 @@ export default {
...
+        this.player.currentTime < this.currentItem.time &&
+        this.player.currentTime >= this.currentItem.time - POP_UP_CHECKING_FREQUENCY
+      )
+        timeToInspect += POP_UP_CHECKING_FREQUENCY;
+
+      const firstUnansweredInteraction = this.checkMovingToTimestampAllowed(
+        timeToInspect
+      );
+      if (firstUnansweredInteraction != null) {
+        this.setPlayerTime(firstUnansweredInteraction.time - POP_UP_CHECKING_FREQUENCY);
```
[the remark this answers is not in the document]
dalmia: because our check for whether item should be shown or not checks for a window of POP_UP_CHECKING_FREQUENCY **before** the actual timestamp. Probably doing `-0.1` or something should also work but would be more unintuitive. should I add a comment here?

dalmia:
> There are two places where the tooltip comes on top of the active div in touch/mobile. One is when I click the Preview button. Once is when I click Publish button. Can you fix this in this PR itself? Should be a small change

> Also, a separate bug, not for this PR. In subjective question textbox, if I keep typing something without any space, the submit button stays hidden idk why. Only if I press spacebar or close the mobile keyboard, the submit button gets active

made a separate issue for these two - any of us can solve them in a separate PR - don't want to further clutter this one.

dalmia:
> I had this weird on mobile phone where in preview mode and in player as well, the questions were not getting updated to green color after answering. I guess this is not because of this PR, but due to this, I was being redirected to the same question again and again because the code thought that no question has been answered.

preview mode mein we are not recording the question ka answers kabhi. toh whenever you close preview mode, the answers are reset. no session is created for preview mode ka responses na? player ke saath toh nahi hona chaiye - can you actually check if that is happening again?

dalmia:
> What should we do for cases where, let's say someone has answered a plio where skip was enabled. They answered some questions and skipped some questions. Now, the creator changed the setting to "skip enabled" as false. Now when the user returns to the plio, the code will drop the user at the last unanswered question, even if there are some other unanswered questions in between. The code is written with the assumption that all questions will be answered in order. So I'm not sure what should the code ideally do in this case.

haan this is a valid concern. instead of restricting them from updating after published, we can simply set the last unanswered index to the last index before any unanswered questions begin - that makes sense? the reason is that I want to really avoid restricting people for edge case scenarios if they can be handled by us in a decent manner.

src/pages/Embeds/Plio.vue
dalmia: videoSeeked wala function mein we are calling `moveToFirstUnansweredItemTimestampOrPass` right? wahan pe ho raha same check!

src/pages/Embeds/Plio.vue
dalmia: haan made a suggestion towards the end of the comments. let me know what you think!

src/pages/Embeds/Plio.vue
dalmia: haan the trick is to decide if the toast is to be shown in a different function than where this check was done - the function that pops up the item - thik hain - will have a look!

src/pages/Embeds/Plio.vue
```diff
@@ -659,7 +686,47 @@ export default {
...
     },
+    checkMovingToTimestampAllowed(timestamp) {
+      if (!this.isSkipEnabled && this.lastAnsweredInteractionIndex < this.numItems - 1) {
+        const firstUnansweredInteraction = this.items[
+          this.lastAnsweredInteractionIndex + 1
+        ];
+        if (timestamp >= firstUnansweredInteraction.time)
+          return firstUnansweredInteraction;
+      }
+    },
```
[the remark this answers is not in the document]
dalmia: haan nice suggestion 👍

src/pages/Embeds/Plio.vue
```diff
@@ -202,6 +209,8 @@ export default {
       isAspectRatioChecked: false, // whether the check for aspect ratio has been done
       watchingEventDBId: null, // the DB id of the latest 'watching' event for a given session
       plioSettings: null, // stores this plio's settings
+      lastAnsweredItemIndex: -1, // index of the interaction that was last answered
```
deepansh96, line 212:
```suggestion
      lastAnsweredItemIndex: -1, // index of the item that was last answered
```

src/pages/Embeds/Plio.vue
```diff
@@ -656,12 +700,53 @@ export default {
...
         let maximizeButton = document.getElementById("plioMaximizeButton");
-        if (maximizeButton != undefined) this.mountOnFullscreenPlyr(maximizeButton);
+        if (maximizeButton != undefined)
+          this.mountOnFullscreenPlyr(maximizeButton);
       });
     },
+    isMovingToTimestampAllowed(timestamp) {
+      if (this.firstUnansweredItem == null) return true;
+      // check whether skip is disabled and all items have been answered
+      return timestamp < this.firstUnansweredItem.time;
```
deepansh96, line 710: here the comment says that `check whether skip is disabled...` but that check is not happening in the code?

src/pages/Embeds/Plio.vue
```diff
@@ -656,12 +700,53 @@ export default {
...
+    },
+    /**
+     * whether the user should be moved to the first unanswered item
+     * or allowed to pass through to whatever timestamp they are trying to
```
deepansh96, line 714:
```suggestion
     * or is allowed to pass through to whatever timestamp they are trying to
```

src/pages/Embeds/Plio.vue
```diff
@@ -656,12 +700,53 @@ export default {
       // insert the button inside the plyr instance so that it shows up in fullscreen mode
       this.$nextTick(() => {
         let maximizeButton = document.getElementById("plioMaximizeButton");
-        if (maximizeButton != undefined) this.mountOnFullscreenPlyr(maximizeButton);
+        if (maximizeButton != undefined)
+          this.mountOnFullscreenPlyr(maximizeButton);
       });
     },
+    isMovingToTimestampAllowed(timestamp) {
```
deepansh96, line 707: add @param for timestamp?

src/pages/Embeds/Plio.vue
```diff
@@ -511,7 +544,13 @@ export default {
   },
   methods: {
     ...mapActions("auth", ["setAccessToken", "setActiveWorkspace"]),
-
+    isItemResponseEmpty(itemIndex, answer) {
```
deepansh96, line 547: add @param for `itemIndex` and `answer`

src/pages/Embeds/Plio.vue
```diff
@@ -656,12 +700,53 @@ export default {
...
+      if (this.isSkipEnabled) return false;
+      let timeToInspect = this.player.currentTime;
+
+      /**
+       * this is required to handle the case that the user toggles
+       * the show video/question button while an item is active and tries
+       * to play the video since the user will be in the time range of item
+       * timestamp and item timestamp - POP_UP_CHECKING_FREQUENCY, the code
+       * after this snippet will function incorrectly
+       */
```
deepansh96, line 727:
```suggestion
      /**
       * this is required to handle the case when the user toggles
       * the show video/question button while an item is active and tries
       * to play the video. 
       * since the user will be in the time range of item timestamp and 
       * item timestamp - POP_UP_CHECKING_FREQUENCY, the code
       * after this snippet will function incorrectly
       */
```

deepansh96 (changes requested):

deepansh96 (approved):

PR #659, opened 2025-05-03
Fix workspace URL inconsistency after logout and login

### LINKED ISSUE #459: Logging out when in org workspace and logging back in causes inconsistency
steps to reproduce:
- log in to an account with >= 2 workspaces
- switch to org workspace
- log out
- log back in
- you should see yourself in the org workspace but the link would be `/home` instead of `/af/home`.

G4URAV001:
Fixes #459

## Summary

This PR fixes the workspace URL inconsistency issue that occurs when logging out from an org workspace and logging back in.

**Issue:**
When a user logs out from an org workspace and logs back in, they would see themselves in the org workspace but the URL would show `/home` instead of `/<workspace>/home`.

**Root Cause:**
The issue was in the logout process:
- The application was redirecting to the login page **before** clearing the active workspace from the store
- Due to this sequence, the workspace information was still present in the store during the redirect
- When logging back in, the application correctly restored the workspace in the state, but the URL path wasn't properly updated to include the workspace shortcode

**Fix:**
Modified the `logoutUser()` function in `App.vue` to:
1. Clear the active workspace state **before** initiating the redirect to the login page
2. Remove the redundant workspace clearing that was happening after the redirect

## Test Plan

- [x] Manually verified the fix by following the steps to reproduce:
  1. Logged in to an account with ≥ 2 workspaces
  2. Switched to org workspace
  3. Logged out
  4. Logged back in
  5. Confirmed that the URL correctly shows the workspace shortcode (e.g., `/af/home` instead of just `/home`)

- [x] Test Responsiveness
   - [x] Laptop (1200px)
   - [x] Tablet (760px)
   - [x] Phone (320px)
- [x] Cross-Browser Testing
   - [x] Chrome
   - [x] Firefox
- [x] Local Language Support
   - [x] Self-review
   - [x] Check for bundle size [here](https://bundlephobia.com/) if adding a package (N/A - no new packages)
   - [ ] If adding or removing any environment variable:
       - [ ] update `docs/ENV.md`
       - [ ] update DockerFile
       (N/A - no env variable changes)
- [x] Testing
   - [ ] Wrote tests (N/A - simple fix for existing functionality)
   - [x] Tested locally
   - [ ] Tested on staging
   - [ ] Tested on production
- [x] Lighthouse Checks (N/A - no UI changes)
   - [x] Images have `alt` attributes
   - [x] Any `` tags have `width` and `height` specified
   - [x] Any `target="_blank"` links have `rel="noopener"`
   - [x] Only SVGs are used as images. If PNGs are used, their size has been optimised.
   - [x] Any SVGs without text have their `aria-label` attributes set

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -804,6 +804,9 @@ export default {
     },
     /** logs out the user */
     logoutUser() {
+      // First clear the active workspace to prevent it from being used during redirect
+      this.unsetActiveWorkspace();
+      
       this.unsetAccessToken().then(() => {
         this.$router.replace({
           name: "Login",
@@ -816,9 +819,6 @@ export default {
         // added here so that if someone clicks on logout while
         // some activity is pending
         this.stopLoading();
-
-        // clear active workspace
-        this.unsetActiveWorkspace();
       });
     },
     onClose(event) {


G4URAV001:
cc: @dalmia

PR #664, opened 2026-04-05
fix: dialog language not updating on language switch (#517)

### LINKED ISSUE #517: Changing language from the switcher does not update the language in a dialog box

santhosh-7777:
Fixes #517
###  Problem
Dialog titles, descriptions, and button texts were stored as translated strings using `$t()`.
Because of this, they did not update when the language was changed.
###  Reproduction
1. Open editor.
2. Click "Publish".
3. Change language.
4. Dialog text does not update.
### Solution
- Store translation keys instead of translated strings.
- Remove `$t()` from dialog configuration.
- Ensure translations are applied reactively.
### Result
Dialog content now updates correctly when the language is switched.
### Testing
- Tested locally
- Verified dialog updates after language change


santhosh-7777:
@rathorevaibhav  Could you please review when you get time?

